PR #3, opened 2018-01-24
Improve docker test commands

jcbwlkr:
+ Use 12 character IDs instead of 3. In testing I ran into collisions.
+ Don't need to `cmd.Wait` after `cmd.Run`. Since Run calls Wait.
+ Remove unnecessary string conversion
+ Prefer structure for `docker inspect` output over type assertions

### DIFF
--- a/internal/platform/docker/docker.go
+++ b/internal/platform/docker/docker.go
@@ -23,28 +23,32 @@ func StartMongo() (*Container, error) {
 		return nil, fmt.Errorf("starting container: %v", err)
 	}
 
-	id := out.String()[:3]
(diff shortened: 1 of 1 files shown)


wyattjoh:
Any reason why you aren't using the docker Go library to orchestrate?

jcbwlkr:
@wyattjoh that's a good question but I can't really answer it. For this PR I'm mostly just trying to improve the existing version when I found it.

ardan-bkennedy:
If this simple code works and we don't need the dependency, let's not use it.

PR #13, opened 2018-06-27
logging changes

ardan-bkennedy:
Changes to remove the use of the stdlib singleton logger.

### DIFF
--- a/cmd/crud/handlers/health.go
+++ b/cmd/crud/handlers/health.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"log"
 	"net/http"
 
 	"github.com/ardanlabs/service/internal/platform/db"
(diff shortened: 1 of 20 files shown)


cmd/crud/handlers/health.go
```diff
@@ -11,11 +12,12 @@ import (
 
 // Health provides support for orchestration health checks.
 type Health struct {
+	Log      *log.Logger
```
jcbwlkr: This struct field is unused since the logger is passed in to each handler. We should either pass it in to the handlers or have it as a struct field like this but not both.

cmd/crud/handlers/user.go
```diff
@@ -28,13 +29,14 @@ func check(err error) error {
 
 // User represents the User API method handler set.
 type User struct {
+	Log      *log.Logger
```
jcbwlkr: This field is also unused, same as on `type Health`.

cmd/crud/handlers/health.go
ardan-bkennedy: Excellent Catch!!

cmd/crud/handlers/user.go
ardan-bkennedy: Brilliant

ardan-bkennedy:
I think `log` is the best name for the variable. Using `l` is too short and anything else is an attempt to not use `log`. Let's keep it.

cmd/crud/handlers/health.go
nstogner: Might not be applicable to this PR but: Why pass DB handle via struct and logger via arg?

nstogner: Ahh, I see b/c all handlers need a logger? And only some need DB handle?

nstogner:
Looks good on 1st pass

Issue #16, opened 2018-06-27
Consider using a struct for config


nstogner:
I like the idea of having a `type config struct` in the `main` package. It provides a simple declarative way of showing all inputs that can effective application behavior. You can then populate it from the env (i.e. https://github.com/kelseyhightower/envconfig) or somewhere else on startup.

nstogner:
This also makes it easy to inspect configuration options/defaults. For example: https://godoc.org/github.com/kelseyhightower/envconfig#Usage

ardan-bkennedy:
I have a philosophy problem with this. I think it hides information. I don't want to pass all the configuration around, especially when it is not all needed in all places. I think what we are doing is solving the problem you descibed.

nstogner:
I wouldn't advocate for passing the entire struct to things created from `main()`, just the variables that they need
```
func main() {
    var cfg struct{
        ABC string
        XYZ int
    }

    // TODO: Process config from source (i.e. env)

    something := NewSomething(cfg.ABC, cfg.XYZ)
}
```

ardan-bkennedy:
So is this really enhancing readability of the code or just adding to it?

ardan-bkennedy:
Nick, create a PR for this and let's look at it. I think it is just adding an extra local type that is not necessary but maybe it will enhance readability.

ardan-bkennedy:
Here is the pull request for this change.

https://github.com/ardanlabs/service/pull/22

ardan-bkennedy:
This has been merged.

PR #20, opened 2018-06-27
Use declarative configuration struct in main

nstogner:
Example for #16. This PR uses an anonymous type in `main` as a form of making configuration more declarative. IMO this makes configuration inputs more searchable. Possible benefits include: easy exposure of configuration via a future API endpoint along with the ability to print configuration usage. The `envconfig` package is used as reference implementation... it would probably be replaced with a kit package.


ardan-bkennedy:
I absolutely don't want to add another dependency for this. I rather add support our cfg package.

ardan-bkennedy:
Shit, that looks pretty nice! So that can replace our cfg package. Then we have the issue of adding cli flags.

ardan-bkennedy:
Remove the cfg package from platform

ardan-bkennedy:
Let's drop the PR and re-create it.

ardan-bkennedy:
By using this new package, we lost our ability to have the config package create the flags. So I guess we need a new package that can take a config struct value and build the command line arguments?

wyattjoh:
I know you're opposed to new packages... But this one does pretty much what you want @ardan-bkennedy: https://github.com/urfave/cli#flags

ardan-bkennedy:
This repo has a lot of code for this. There are also two other packages it maintains :(

ardan-bkennedy:
We must assume defaults are in use and the variables are not defined. It's the struct that defines the configuration points and defaults. It is the point of truth.

ardan-bkennedy:
I will be closing this PR because of PR 22

https://github.com/ardanlabs/service/pull/22

PR #22, opened 2018-06-28
config changes

ardan-bkennedy:
Changes to use envconfig


jcbwlkr:
Looks good to me. Needs an explicit dependency constraint added to Gokpg.toml though

ardan-bkennedy:
The `dep1` tool didn't do that?

ardan-bkennedy:
Added the constraint and updated vgo support.

Issue #18, opened 2018-06-27
Authentication


ardan-bkennedy:
We need to add a minimal but complete layer for authentication and authorization.

nstogner:
First thoughts are to use JWT and add https://github.com/dgrijalva/jwt-go as a dependency to the middleware package. Tokens can be generated elsewhere.

ardan-bkennedy:
We should teach to how to maintain secrets in code as well.

nstogner:
I think we should use asymmetric encryption (RSA). This way the service that consumes/validates tokens only needs the RSA public key which mitigates secret exposure risk. A separate service would be used for token generation. In this service we would demonstrate the handling of secrets (RSA private key) and possibly the storage of user passwords. This token-generation service would most likely manage the CRUD for user resources.

For reference:
https://auth0.com/blog/json-web-token-signing-algorithms-overview/

nstogner:
@ardan-bkennedy Do we want to support multi-tenancy? My vote is no for simplicity's sake.

jcbwlkr:
I vote no on multi tenancy

ardan-bkennedy:
I vote no as well.

nstogner:
Closed, will open Authorization issue separately.

PR #23, opened 2018-06-29
Add JWT auth middleware

nstogner:
Currently a WIP... Adding middleware for auth via JWTs. Addresses #18.


jcbwlkr:
Some initial thoughts:

1. I like the `dgrijalva` package so +1 on using that.

2. Could we put more of the actual work in the `internal/auth` package and not in the middleware itself? Maybe a function in `auth` like

```func Parse(token string) (Claims, error)```

Then the middleware just handles translating from HTTP headers and passing along the claims.

3. What is the thought around having a KeyFunc and making the claims include a key ID? What if we keep it simple and the server just has one Public/Private key pair?

**Comment by nstogner** [comment]
2. That works

**Comment by nstogner** [comment]
For point 3: This allows for the seamless changing of tokens. To show a simple hardcoded value we could pass in `func(kid string) (*rsa.PublicKey, error) { return <hardcoded-key>, nil }`. I think it also provides a talking point / framework for how to extend this in production. IMO there are countless examples of MVP services in the wild that do not demonstrate how to accomplish things like this in production. I think there is little value to demonstrate another not-production-ready service. This is a larger point we should probably discuss.

**Comment by ardan-bkennedy** [comment]
I'm impressed you can do a review here with all the files that changes. Could we schedule a code review with vscode live?

File: internal/auth/claims.go [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+	RoleTODO  = "TODO:MORE_ROLES"
+)
+
+// signingAlg is the agreed-upon algorithm that this application uses to sign
+// and validate tokens.
+//
+// The method used to sign the JWT must be validated to avoid a critical
+// vulnerability:
+// https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/
+var signingAlg = jwt.SigningMethodRS256
```
ardan-bkennedy: Can we remove the Global somehow?

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Claims represents the authorization claims transmitted via a JWT.
+type Claims struct {
+	Roles []string `json:"roles"`
+	jwt.StandardClaims
+}
+
+// GenerateToken generates a JWT token string.
+func GenerateToken(key *rsa.PrivateKey, kid string, claims Claims) (string, error) {
```
ardan-bkennedy: For when I have to teach, what does `kid` stand for?

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+	tkn.Header["kid"] = kid
+	return tkn.SignedString(key)
+}
+
+// KeyFunc is used to map a JWT key id (kid) to the corresponding public key.
+// USE CASE:
+// Private keys should be rotated. During the transition
+// period, tokens signed with the old and new keys can coexist by looking up
+// the correct public key by key id (kid).
+//
```
ardan-bkennedy: This space in the comment is bothering me. Maybe use a `*` for each point?

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+// Key-id-to-public-key resolution is usually accomplished via a public JWKS
+// endpoint. See https://auth0.com/docs/jwks for more details.
+type KeyFunc func(kid string) (*rsa.PublicKey, error)
+
+// NewParser is the factory function for a Parser.
+func NewParser(kf KeyFunc) *Parser {
```
ardan-bkennedy: This is where you should pass the Signing Algorithm.

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+// Parser wraps jwt.Parser with the ability to fetch keys based on kid.
+type Parser struct {
+	kf KeyFunc
+	p  *jwt.Parser
+}
+
+// ParseClaims from a token string.
+func (p *Parser) ParseClaims(tknStr string) (Claims, error) {
+	var claims Claims
+
```
ardan-bkennedy: Remove the space for readability. It is used by the next line of code. I almost asked you to remove it because it was not obvious at first.

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+	tkn, err := jwt.ParseWithClaims(tknStr, &claims, func(t *jwt.Token) (interface{}, error) {
```
ardan-bkennedy: Please put the literal function above and in its own variable `f`. I can never read code like this and find it super clever. Plus, the literal function is so much easier to read when it is on its own.

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+		kid, ok := t.Header["kid"]
```
ardan-bkennedy: This is interesting. `t` is a parameter to the literal function. Where does it come from? Is there anyway to know that?

internal/auth/claims.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		return p.kf(kidStr)
+	})
+	if err != nil {
+		return Claims{}, err
+	}
+
+	if !tkn.Valid {
+		return Claims{}, errors.New("Invalid token")
```
ardan-bkennedy: Just noticed this is an `internal` package and not `platform`. This should be wrapping errors in all cases.

internal/auth/claims_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"crypto/rsa"
+	"fmt"
+	"testing"
+
+	"github.com/ardanlabs/service/internal/auth"
+	jwt "github.com/dgrijalva/jwt-go"
+)
+
+func TestParseClaims(t *testing.T) {
+	// Parse the private key used to generate the token.
```
ardan-bkennedy: Space after comments for readability.

internal/auth/claims_test.go
```diff
@@ -0,0 +1,103 @@
...
+	// Assert expected claims.
+	if exp, got := len(signedClaims.Roles), len(parsedClaims.Roles); exp != got {
+		t.Fatalf("expected %v roles, got %v", exp, got)
+	}
+	if exp, got := signedClaims.Roles[0], parsedClaims.Roles[0]; exp != got {
+		t.Fatalf("expected roles[0] == %v, got %v", exp, got)
+	}
+}
+
+func singleKeyFunc(id string, key *rsa.PublicKey) auth.KeyFunc {
```
ardan-bkennedy: Unless it is used more than once, I could be missing that, why not just assign this to a variable in the test and use it.

internal/mid/auth.go
```diff
@@ -0,0 +1,58 @@
...
+		return next(ctx, log, w, r, params)
+	}
+
+	return h
+}
+
+// getBearerToken grabs a token from the request header. Expected header is of
+// the format `Authorization: Bearer <token>`.
+func getBearerToken(r *http.Request) (string, error) {
+	hdr := r.Header.Get("Authorization")
```
ardan-bkennedy: This is where we get into fun conversations about not hiding information and testability. We could pass the "Authorization" string into this function and make it more precise. It's the only thing we need from the request. Have the calling function do this. The calling function Needs the request so it makes more sense there . IMO.

This also makes the function more testable.

internal/auth/claims.go
```diff
@@ -0,0 +1,82 @@
...
+	kf KeyFunc
+	p  *jwt.Parser
+}
+
+// NewParser is the factory function for a Parser.
+func NewParser(kf KeyFunc, validAlgNames []string) *Parser {
+	return &Parser{
+		kf: kf,
+		p: &jwt.Parser{
+			// The algorithm used to sign the JWT must be validated to avoid a critical
```
ardan-bkennedy: Line space before comments for readability.

internal/auth/claims.go
```diff
@@ -0,0 +1,82 @@
...
+			return nil, errors.New("Missing key id (kid) in token header")
+		}
+		kidStr, ok := kid.(string)
+		if !ok {
+			return nil, errors.New("Token key id (kid) must be string")
+		}
+
+		return p.kf(kidStr)
+	}
+	var claims Claims
```
ardan-bkennedy: Hum... My first instinct is a line space but I get why you didn't. @jcbwlkr, please chime in. I think I always line space here.

internal/mid/auth.go
```diff
@@ -0,0 +1,57 @@
...
+
+		return next(ctx, log, w, r, params)
+	}
+
+	return h
+}
+
+// getBearerToken grabs a token from the request header. Expected header is of
+// the format `Authorization: Bearer <token>`.
+func getBearerToken(hdr string) (string, error) {
```
ardan-bkennedy: Should the parameter be called auth or something similar? hdr doesn't tell me anything.

internal/auth/claims.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, errors.New("Token key id (kid) must be string")
+		}
+
+		return p.kf(kidStr)
+	}
+
+	var claims Claims
+	tkn, err := jwt.ParseWithClaims(tknStr, &claims, f)
+	if err != nil {
+		return Claims{}, err
```
ardan-bkennedy: wrap error here

ardan-bkennedy:
@jcbwlkr I am good with the code readability review. I don't understand the mechanics/semantics enough.

PR #25, opened 2018-07-07
Flags

ardan-bkennedy:
> 1. The branch adds `spf13/pflag` but doesn't use it. Presumably that's from an earlier iteration but if we're not using it then remove it from dep and vendor.

I've removed it from vendor. Great catch!

> 2. The code panics if you run a program without providing any arguments. Line 18 says `if osArgs[1] == "-h" || osArgs[1] == "--help" {` but the args slice is not guaranteed to have that many elements.

I had not finished the error handling but great catch. I wanted review on the API. Fixed.

> 3. When you run print the usage of a stdlib `flag` package it starts with a header line like `Usage of myApp` where "myApp" comes from `os.Args[0]`. Would be nice to see here.

I was thinking about it but again it was secondary. I fixed it! Good catch.

> 4. Need to add support for more types like bool and the other numeric types besides `int`.

Yes, I am only handling what we need right now. I rather add support for more types as we need it. Let's not add more code than we need.

> 5. Flag parsing should stop when it sees a stand alone `--`. Meaning if I do `./binary --timeout 5s` then it sets the timeout flag to 5 seconds but `./binary -- --timeout 5s` it should treat everything after the `--` as just arguments and not flags.

Yea, error handling and great catch. I fixed it.

> 6. On that note there should be an easy way to get the arguments once flag parsing is done.

Not sure I understand but other things have been fixed and there are tests.

> All of these things are already solved by `flag` or `spf13/pflag`. Maybe we could just make our package be a thin wrapper around those packages that lets us do the convenient "here's a struct, process it" kind of code.

Yes, those things are but that package is the wrong package for what we need. I want to stress again, we will never manually create flags. Everything is driven off the config value. I rather keep this package focused and clean towards that idea.

> I guess really I am uneasy about this package in general. Why are we writing it? Why not use a well established package?

I did look for something we could use. Asked others as well. I didn't find anything and we have the skill set to do this. So now we have a package that does what we need and is minimal.


cmd/crud/main.go
```diff
@@ -58,21 +58,34 @@ func main() {
...
+
+	log.Println("main : Started : Application Initializing")
+	defer log.Println("main : Completed")
+
 	cfgJSON, err := json.MarshalIndent(cfg, "", "    ")
 	if err != nil {
 		log.Fatalf("main : Marshalling Config to JSON : %v", err)
 	}
-	log.Printf("config : %v\n", string(cfgJSON))
+	log.Printf("main : Config : %v\n", string(cfgJSON))
```
wyattjoh: Is it ok to print the configuration to the logs? This essentially leaks credentials to a potentially unprotected log file.

ardan-bkennedy: UGH!  We fixed this with our Config package by filtering certain keywords. Now this has moved back up to the app level. GREAT CATCH. All we can do it add a TODO, to check what config is being written?

PR #30, opened 2018-07-12
changing time to a dependency for create and update. Closes #4

### LINKED ISSUE #4: Consider passing in time.Time as a argument to functions using time.Now()
Consider removing `time.Now()` and passing in a `now time.Time` parameter into functions such as `internal/user.Create(...)` so that time is treated as a dependency.

https://github.com/ardanlabs/service/blob/master/internal/user/user.go#L72

Zach-Johnson:


nstogner:
Thanks for the PR! @jcbwlkr are you happy with this?

ardan-bkennedy:
LGTM!

PR #35, opened 2018-08-07
Dockerfile best practises.

johnharris85:
- Cleaned up Dockerfiles for current best practise
- Refactored makefile for new Dockerfiles / performance

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -15,7 +15,7 @@ services:
     container_name: mongo
     networks:
       - shared-network
-    image: mongo
+    image: mongo:3-jessie
     ports:
       - 27017:27017
     command: --bind_ip 0.0.0.0
@@ -25,7 +25,7 @@ services:
     container_name: crud
     networks:
       - shared-network
-    image: crud-amd64
+    image: crud-amd64:1.0
     ports:
       - 3000:3000 # CRUD API
       - 4000:4000 # DEBUG API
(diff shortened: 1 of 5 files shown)


docker-compose.yaml
```diff
@@ -15,7 +15,7 @@ services:
     container_name: mongo
     networks:
       - shared-network
-    image: mongo
+    image: mongo:3-jessie
```
ardan-bkennedy: I understand everything but this change?

makefile
```diff
@@ -3,22 +3,31 @@ SHELL := /bin/bash
 all: crud metrics tracer
 
 crud:
-	export GO111MODULE=on
 	cd "$$GOPATH/src/github.com/ardanlabs/service"
-	docker build -t crud-amd64 -f dockerfile.crud .
-	docker system prune -f
```
ardan-bkennedy: I want to run prune after the build automatically. It helps to keep my system clean.

docker-compose.yaml
johnharris85: Best practice to pin images rather than relying on `latest` (same reason as with code dependencies). It's even better to pin the image SHA to guarantee version, but went with this more 'user friendly' intermediate step first.

makefile
johnharris85: The problem with `system` pruning after every build is that you destroy the build cache. What artifacts are you concerned about cleaning up? (there are different prune scopes beside system that might give a better result for what you want).

ardan-bkennedy:
I might build a lot during a dev cycle so I don’t care about the cache. The build is still fast. The artifacts are eating a lot of disk space. So I rather take the cost of pruning and not worry my disk gets full.

PR #36, opened 2018-08-09
Fixing TraceIDHeader and replace deprecated func

farrukhny:
I think TraceIDHeader should set only trace id and it is returning extra info and value on diff format 

screen shot 2018-08-09 at 9 51 51 am

Looks like NewSpanWithRemoteParent func has been deprecated and recommend using StartSpanWithRemoteParent.

### DIFF
--- a/internal/platform/web/web.go
+++ b/internal/platform/web/web.go
@@ -78,10 +78,7 @@ func (a *App) Handle(verb, path string, handler Handler, mw ...Middleware) {
 		// Set the parent span on the outgoing requests before any other header to
 		// ensure that the trace is ALWAYS added to the request regardless of
 		// any error occuring or not.
-		data, err := trace.MarshalSpanContext(span.SpanContext())
-		if err == nil {
-			w.Header().Set(TraceIDHeader, string(data))
-		}
+		w.Header().Set(TraceIDHeader, v.TraceID)
 
 		// Call the wrapped handler functions.
 		if err := handler(ctx, a.log, w, r, params); err != nil {
--- a/internal/platform/trace/trace.go
+++ b/internal/platform/trace/trace.go
@@ -25,8 +25,7 @@ func WithSpanContext(ctx context.Context, name string, spanContext string) (cont
 			return ctx, span
 		}
 
-		span := trace.NewSpanWithRemoteParent(name, sc, trace.StartOptions{})
-		ctx := trace.WithSpan(ctx, span)
+		ctx, span := trace.StartSpanWithRemoteParent(ctx, name, sc)
 		return ctx, span
 	}
 


internal/platform/web/web.go
```diff
@@ -78,10 +78,7 @@ func (a *App) Handle(verb, path string, handler Handler, mw ...Middleware) {
 		// Set the parent span on the outgoing requests before any other header to
 		// ensure that the trace is ALWAYS added to the request regardless of
 		// any error occuring or not.
-		data, err := trace.MarshalSpanContext(span.SpanContext())
-		if err == nil {
-			w.Header().Set(TraceIDHeader, string(data))
-		}
+		w.Header().Set(TraceIDHeader, v.TraceID)
```
ardan-bkennedy: This is going to hurt me on the sidecars. Let's talk. I am trying to make sure there is a single picture of trace across process boundaries. I could not find any other way to do it.

ardan-bkennedy: Look at these lines of code. How do you do this with just a trace id. I couldn't figure it out

https://github.com/ardanlabs/service/blob/master/internal/platform/web/web.go#L64

farrukhny: I think you already checking for existing trace on your WithSpanContext https://github.com/ardanlabs/service/blob/master/internal/platform/trace/trace.go#L21

and also you are checking same span.SpanContext which is returned from WithSpanContext and in that case i think value always should be same.

https://github.com/ardanlabs/service/blob/master/internal/platform/web/web.go#L81

farrukhny: Or maybe we can create separate header fields for SpinID and TraceOptions if you need to pass it to sidecars

ardan-bkennedy: When a new span is created in process B, it needs to belong to the same trace that was started in process A. It looks like this concept is missing in the API, which is why I am marshaling that entire value. Doing this I got it to work.

Maybe there is something new in the API that supports this. Or, maybe I am confused and missing something.

ardan-bkennedy: @farrukhny we should finalize this issue sometime this week.

farrukhny: @ardan-bkennedy I am not sure what the issue maybe I am missing something but I did a test of the PR and looks like everything working, please take a look at screenshots

screen shot 2018-08-13 at 10 29 13 am


screen shot 2018-08-13 at 10 35 15 am


screen shot 2018-08-13 at 10 35 36 am

ardan-bkennedy: Hey @farrukhny when can we get on a call. I have no doubt I'm doing something wrong. But I think we need to talk.

farrukhny: @ardan-bkennedy sure, I will send my number on twitter

ardan-bkennedy: the changes in trace.go are probably important since you identified we are using a deprecated function now. But after our conversation we don't want to make the other changes. So can you rollback the web.go changes and just keep the trace.go changes. @farrukhny

farrukhny: Sure, will do that

PR #44, opened 2018-09-06
WIP: Add gentkn cli command

nstogner:
**WORK IN PROGRESS**

Add a cli tool for generating JWT's.

### DIFF
--- a/cmd/gentkn/main.go
+++ b/cmd/gentkn/main.go
@@ -40,11 +40,12 @@ func main() {
 	}
 	*/
 
-	// Source the private key.
-	var keyRdr io.Reader
 	if len(os.Args) < 2 {
 		log.Fatal("main : Missing Required Param <key-file>")
 	}
+
+	// Source the private key.
+	var keyRdr io.Reader
 	switch file := os.Args[1]; file {
 	case "-":
 		keyRdr = os.Stdin
@@ -66,7 +67,6 @@ func main() {
 		log.Fatalf("main : Parsing RSA Private Key : %v", err)
 	}
 
-	// Build the claims.
 	now := time.Now()
 	claims := auth.Claims{
 		StandardClaims: jwt.StandardClaims{
@@ -82,7 +82,6 @@ func main() {
 		log.Fatalf("main : Invalid Claims : %v", err)
 	}
 
-	// Generate and sign the token.
 	tkn, err := auth.GenerateToken(cfg.KeyID, key, jwt.SigningMethodRS256, claims)
 	if err != nil {
 		log.Fatalf("main : Generating Token : %v", err)
--- a/cmd/gentkn/main.go
+++ b/cmd/gentkn/main.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"time"
+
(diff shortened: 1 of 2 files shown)


cmd/gentkn/main.go
```diff
@@ -0,0 +1,91 @@
...
+	var cfg struct {
+		KeyID string `envconfig:"KEY_ID" default:"default"`
+
+		Claims struct {
+			Audience string   `envconfig:"AUDIENCE"`
+			Subject  string   `envconfig:"SUBJECT"`
+			Roles    []string `envconfig:"ROLES"`
+		} `envconfig:"CLAIMS"`
+	}
+	if err := envconfig.Process("GENTKN", &cfg); err != nil {
```
ardan-bkennedy: Why are we not using our internal flag package? Then you also get the command line options for free. If there are bugs let's fix them. I know I have one pending.

cmd/gentkn/main.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := flag.Process(&cfg); err != nil {
+		if err != flag.ErrHelp {
+			log.Fatalf("main : Parsing Command Line : %v", err)
+		}
+		return // We displayed help.
+	}
+	*/
+
+	// Source the private key.
+	var keyRdr io.Reader
```
ardan-bkennedy: This variable declaration should move down.

cmd/gentkn/main.go
```diff
@@ -0,0 +1,91 @@
...
+	if len(os.Args) < 2 {
+		log.Fatal("main : Missing Required Param <key-file>")
+	}
+	switch file := os.Args[1]; file {
```
ardan-bkennedy: I think we need a line feed here and it's probably a good place for that variable.

cmd/gentkn/main.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		log.Fatalf("main : Reading Key : %v", err)
+	}
+
+	key, err := jwt.ParseRSAPrivateKeyFromPEM(keyContents)
+	if err != nil {
+		log.Fatalf("main : Parsing RSA Private Key : %v", err)
+	}
+
+	// Build the claims.
```
ardan-bkennedy: Not sure this comment is adding value.

cmd/gentkn/main.go
```diff
@@ -0,0 +1,91 @@
...
+			IssuedAt:  now.Unix(),
+			Subject:   cfg.Claims.Subject,
+		},
+		Roles: cfg.Claims.Roles,
+	}
+	if err := claims.Valid(); err != nil {
+		log.Fatalf("main : Invalid Claims : %v", err)
+	}
+
+	// Generate and sign the token.
```
ardan-bkennedy: Not sure the comment is adding value.

cmd/gentkn/main.go
ardan-bkennedy: I get that, I want to know why we are passing, non-flagged arguments?

nstogner: Seems to go against unix conventions to have required flags for cli tools

nstogner: That coupled with not really having a good default for that value, seems proper to use a non-flag arg

jcbwlkr:
Closing per conversation with @nstogner. The new PR that adds a token endpoint is preferred

PR #45, opened 2018-09-10
Move to a Garage Sale API theme

### LINKED ISSUE #39: Pick a theme
The material would be easier to digest if our main service wasn't simply `crud` and if we were manipulating records besides just `User`. Some ideas include:

- A TODO service where we are manipulating `Task` values.
- A veterinarian's office service where we have `Pet` values.
- A comic collector's database. Let people keep track of their `Comic`s.
- A garage sale manager. I want to know what `Product`s I have for sale in my garage, how much they cost, and who they came from. Many people do multi-family sales and it could be useful to know that item 12345 came from Cindy so the money should go to her.

We could probably come up with several and bikeshed this forever.

jcbwlkr:
Resolves #39 

To achieve the goal of migrating to a specific theme this PR primarily does the following:

1. The service `crud` is renamed to `sales-api`. This gives a better idea of what purpose the service is intending to serve. Environment variables are updated (`SALES_DB_HOST` vs `CRUD_DB_HOST` etc) as well as all makefile/dockerfiles etc.
2. Adds an internal package and endpoints around a new entity called `Product`. There is a model diagram added to the README.

The implementation of the Product endpoints are very similar to the existing User code but there are some differences I would like to discuss.

1. I am using distinct types `NewProduct` and `UpdateProduct` for POST and PUT operations respectively. I never liked having a single type like `CreateUser` what was used for both creating and updating.
2. I am allowing partial updates on `UpdateProduct`. This is more in line with most of the APIs I have written though it makes the code more complicated. To achieve this I use pointer types like `*string` and `*int`. Normally these are not desirable but as we say in class the time when we tend to make exceptions is around marshalling/unmarshalling. I believe there is sufficient documentation in the code to explain to people why we are doing this.
  2.1. As a side note if this were to really be fully "correct" we would consider doing updates with `PATCH` instead of `PUT` and implementing a spec like `json+patch`. I have done that for production code and it works pretty well but it might be beyond the scope of what we want to do in this class.
3. On API tests I don't want to string match against hard coded JSON. I am using `github.com/google/go-cmp/cmp` instead. I really really like this library for testing.
4. I changed this pattern

```go
if err == mgo.ErrNotFound {
	return ErrNotFound
}
return errors.Wrap(err, fmt.Sprintf("db.products.removeID(%s)", id))
```

With this

```go
if err == mgo.ErrNotFound {
	err = ErrNotFound
}
return errors.Wrap(err, fmt.Sprintf("db.products.removeID(%s)", id))
```

The difference is minor. The value comes when code fails because we can keep the `fmt.Sprintf` context on failures. The important behavior remains the same since we still run the error through `errors.Cause`


ardan-bkennedy:
Not sure I like this
```
if err == mgo.ErrNotFound {
	return ErrNotFound
}
return errors.Wrap(err, fmt.Sprintf("db.products.removeID(%s)", id))
```
With this
```
if err == mgo.ErrNotFound {
	err = ErrNotFound
}
return errors.Wrap(err, fmt.Sprintf("db.products.removeID(%s)", id))
```
I think it goes against readability and the standards of return on the negative path.

PR #46, opened 2018-09-13
Changes for new trace propagation package

ardan-bkennedy:
This is using the trace propagration package from opencensus.


nstogner:
LGTM

PR #47, opened 2018-09-19
Update user patterns and add auth endpoint for getting a token.

jcbwlkr:
This does not yet solve the problem of how operators should generate / store keys but I figured we should review + merge what I have so far before going much further.


cmd/sales-api/handlers/user.go
```diff
@@ -13,9 +14,17 @@ import (
 	"go.opencensus.io/trace"
 )
 
+// UserAuth knows everything needed to authenticate Users.
```
nstogner: This comment was a bit confusing to me ... authenticating a user could refer to two things: authenticating for getting a token or authenticating via validating a token (on another endpoint)

cmd/sales-api/handlers/user.go
```diff
@@ -107,10 +116,33 @@ func (u *User) Delete(ctx context.Context, log *log.Logger, w http.ResponseWrite
...
+
+	dbConn := u.MasterDB.Copy()
+	defer dbConn.Close()
+
+	email, pass, ok := r.BasicAuth()
+	if !ok {
+		return web.ErrUnauthorized
+	}
+
+	tkn, err := user.Authenticate(ctx, dbConn, u.Auth.Key, u.Auth.KeyID, u.Auth.Alg, email, pass)
```
nstogner: The name `Authenticate` for the func was confusing to me due to above comment

cmd/sales-api/main.go
```diff
@@ -62,6 +64,11 @@ func main() {
 			SendInterval time.Duration `default:"15s" envconfig:"SEND_INTERVAL"`
 			SendTimeout  time.Duration `default:"500ms" envconfig:"SEND_TIMEOUT"`
 		}
+		Auth struct {
+			KeyID      string `envconfig:"KEY_ID"`
+			PrivateKey string `envconfig:"PRIVATE_KEY"`
```
nstogner: Small note: I like to name config variables like this "PRIVATE_KEY_FILE" b/c it helps me not wonder whether thats a filename or the contents of the key

internal/user/model.go
```diff
@@ -1,52 +1,48 @@
...
+// changed. It uses pointer fields so we can differentiate between a field that
+// was not provided and a field that was provided as explicitly blank. Normally
+// we do not want to use pointers to basic types but we make exceptions around
+// marshalling/unmarshalling.
+type UpdateUser struct {
+	Name            *string  `json:"name"`
+	Email           *string  `json:"email"` // TODO(jlw) enforce uniqueness.
+	Roles           []string `json:"roles"` // TODO(jlw) ensure only admins can change other users' roles. Ensure only includes valid roles.
+	Password        *string  `json:"password"`
+	PasswordConfirm *string  `json:"password_confirm" validate:"omitempty,eqfield=Password"`
```
nstogner: Is the password confirmation something that should be sent to the API? I have always seen this implemented on the frontend

internal/user/user.go
```diff
@@ -106,21 +106,42 @@ func Create(ctx context.Context, dbConn *db.DB, cu *CreateUser, now time.Time) (
...
+	if !bson.IsObjectIdHex(id) {
+		return ErrInvalidID
 	}
 
-	cu.DateModified = &now
-	for _, cua := range cu.Addresses {
-		cua.DateModified = &now
+	fields := make(bson.M)
+
+	if upd.Name != nil {
```
nstogner: Would it make sense to use something like https://github.com/fatih/structs for these mapping lines?

jcbwlkr: No I don't think so. For this class where we want to keep things simple the value that lib provides wouldn't be worth the complexity it would add in terms of learning a new thing. Plus looking at it I don't see that it handles nil fields the way I want.

cmd/sales-api/main.go
jcbwlkr: Agreed. I'll change it.

cmd/sales-api/handlers/user.go
jcbwlkr: We haven't yet wired up the middleware for validating tokens but I believe this type is how we will pass the public key down to the API so it will eventually mean both kinds of authenticate.

cmd/sales-api/handlers/user.go
jcbwlkr: Hmm I would be open to other names if you have suggestions? The signature of the function includes `email` and `pass` for what it's worth.

internal/user/model.go
jcbwlkr: Yeah I have seen it at the API layer a lot. It isn't critical to the API though and could easily be delegated to the client I guess?

cmd/sales-api/handlers/routes.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // API returns a handler for a set of routes.
-func API(log *log.Logger, masterDB *db.DB) http.Handler {
+func API(log *log.Logger, masterDB *db.DB, userAuth UserAuth) http.Handler {
```
ardan-bkennedy: I am not sure value semantics make sense here? Do we want every function to have it's own copy?

jcbwlkr: Yes because it's config. Maybe it could have a better name. `AuthConfig`?

cmd/sales-api/handlers/user.go
nstogner: Good point, might have had token auth on my mind

internal/user/model.go
nstogner: Gotcha, either way

internal/user/user.go
nstogner: Sounds good!

cmd/sales-api/handlers/user.go
nstogner: I would prefer something like `GetToken` but I may be nitpicking here

cmd/sales-api/handlers/routes.go
nstogner: Would it make more sense to pass in a func or such `func GetUserToken(ctx context.Context, email, pass string) (user.Token, error)` down instead of the underlying private key? One benefit would be that the private key would not be passed down to all handlers, just the ability to make a token.

jcbwlkr: I wouldn't want to pass a function in here. There are three options I would consider:

1. Pass in the config and let each layer pass down what is needed. This is the most explicit and is what we're doing now. This is also what is required by the current implementation of the auth platform package.
2. Redefine the auth package to make token generation a method on a type and have our call stack pass that type all the way down to our business layer. I floated that idea by Bill during a code review and we discussed that since the auth package doesn't need it to be a type+method we would just leave `auth.GenerateToken` as a function.
3. Do the same as option 2 but make this function depend on an interface. I don't see how that would really make the code any better though so I would have to see it implemented to know if I like that path.


In any event I want to delay making that decision right now. My next PR uses this same UserAuth type to wire up the authentication+authorization middlewares. Let's rehash this conversation on that PR when it's up because I think there is some improvement to be had.

jcbwlkr:
@nstogner and @ardan-bkennedy from your comments so far there is only one open issue which I would like to defer to the next PR. If I can get your approval I will merge this one.

nstogner:
I am happy with this!

PR #50, opened 2018-10-10
Refactor/http status errors

nstogner:
Idea: Wrap errors with a status rather than defining `web.ErrNotFound`, etc.

Note: Not a complete PR currently, just a discussion starter right now.

### DIFF
--- a/internal/platform/web/errors.go
+++ b/internal/platform/web/errors.go
@@ -1,20 +1,25 @@
 package web
 
 import (
+	"fmt"
 	"net/http"
 )
 
 type statusError struct {
-	error
+	err    error
 	Status int
 }
 
+func (se *statusError) Error() string {
+	return fmt.Sprintf("http status %v: %s", se.Status, se.err)
+}
+
 func ErrorWithStatus(err error, status int) error {
-	return statusError{err, status}
+	return &statusError{err, status}
 }
 
 func StatusFromError(err error) int {
-	serr, ok := err.(statusError)
+	serr, ok := err.(*statusError)
 	if !ok {
 		return http.StatusInternalServerError
 	}
--- a/internal/platform/web/errors.go
+++ b/internal/platform/web/errors.go
@@ -1,0 +1,28 @@
+package web
+
+import "net/http"
+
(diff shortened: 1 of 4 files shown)


nstogner:
The conceptual question in my head:

Do we need an error variable defined that corresponds to each HTTP status? (`web.ErrNotFound`)
Or is an HTTP status more of an annotation on an error? (`.HTTPStatus()`)

ardan-bkennedy:
let's code review this. The inital idea was to not leak error variables and types outside of the different API layers. Maybe this isn't important in the case of `cmd` level packages.

internal/platform/web/errors.go
```diff
@@ -0,0 +1,28 @@
...
+type HTTPStatuser interface {
+	HTTPStatus() int
+}
+
+type statusError struct {
+	error
+	status int
+}
+
+func (e statusError) HTTPStatus() int {
```
ardan-bkennedy: We need to use pointer semantics here for sure.

ardan-bkennedy: Wait, Maybe not. My brain thought we were implementing the error interface for some reason.

ardan-bkennedy: This is nothing more than a `getter`, so why do we need it. Exposing the `status` field will accomplish the same thing.

ardan-bkennedy: This means we don't need the interface. I don't see a second implementation so honestly, why do we need the decoupling.

ardan-bkennedy: Can't `statusError` implement the `error` interface and we can accomplish the same thing?

nstogner: I see your point on not needing the decoupling the interface provides, a concrete type probably makes sense here.

nstogner:
I agree we should not leak http concepts into our application logic layers... This would just be a different way of communicating http errors within the http layer (`web` pkg / handlers)

ardan-bkennedy:
@nstogner let's try to put this to bed today

jcbwlkr:
Closing this because an improved version is included in `service-training` 👍

PR #52, opened 2018-10-18
Get a real user IP from a request and associating it with a Context

farrukhny:
I am not sure if it is need it for MVP scope if it is not fit to the project feel free to close it.

clientIPFromRequest  functions for extracting real user IP address from a request and associating it with a Context in handler.

### DIFF
--- a/internal/platform/web/web.go
+++ b/internal/platform/web/web.go
@@ -3,7 +3,9 @@ package web
 import (
 	"context"
 	"log"
+	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/dimfeld/httptreemux"
@@ -23,6 +25,7 @@ type Values struct {
 	Now        time.Time
 	StatusCode int
 	Error      bool
+	ClientIP   string
 }
 
 // A Handler is a type that handles an http request within our own little mini
@@ -76,8 +79,9 @@ func (a *App) Handle(verb, path string, handler Handler, mw ...Middleware) {
 		// Set the context with the required values to
 		// process the request.
 		v := Values{
-			TraceID: span.SpanContext().TraceID.String(),
-			Now:     time.Now(),
+			TraceID:  span.SpanContext().TraceID.String(),
+			Now:      time.Now(),
+			ClientIP: clientIPFromRequest(r),
 		}
 		ctx = context.WithValue(ctx, KeyValues, &v)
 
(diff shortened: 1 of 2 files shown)


ardan-bkennedy:
I like these changes. @jcbwlkr and @nstogner what do you think?

internal/platform/web/web.go
```diff
@@ -95,3 +99,25 @@ func (a *App) Handle(verb, path string, handler Handler, mw ...Middleware) {
 	// Add this handler for the specified verb and route.
 	a.TreeMux.Handle(verb, path, h)
 }
+
+// clientIPFromRequest implements effort to return the real client IP
+func clientIPFromRequest(r *http.Request) string {
+	// It parses X-Real-IP and X-Forwarded-For in order to work properly with reverse-proxies such us: nginx or haproxy or AWS ALB/LB
+	// Use X-Forwarded-For before X-Real-Ip as nginx uses X-Real-Ip with the proxy's IP.
+	clientIP := r.Header.Get("X-Forwarded-For")
+	clientIP = strings.TrimSpace(strings.Split(clientIP, ",")[0])
```
nstogner: What is the split has len = 0?

nstogner:
Thanks for submitting the PR! We are always appreciate new requests.

My though process for adding new http functionality like this:

1. Should we be writing common http related code in our service repo or use a library?

I would lean towards using libraries for non-business related code when possible.

2. Does the value we are parsing belong in the context or can we just use a function like `ParseIP(r *http.Request) string, error`

I think we should limit the scope of the context as much as possible, it can quickly become a dumping ground otherwise. A large dependency on context-carried variables becomes hard to reason about.

3. Should we be logging every request that comes through our API?

Hotpath logging is expensive and noisy, should we be doing this by default?

4. Should we be logging PII?

https://www.enterprisetimes.co.uk/2016/10/20/ecj-rules-ip-address-is-pii/

internal/platform/web/web.go
jcbwlkr: I worried about that too but I checked. Split always returns a slice with len >= 1 (unless the substring is blank for some reason but ours isn't, it's `","`).

nstogner: Thanks!

jcbwlkr:
Hi, @farrukhny! Thanks for contributing :)

I am going to close this PR primarily because trusting those proxy headers is only appropriate in some circumstances. Specifically only if the service you are making is being deployed behind a proxy that knows to strip those headers from client requests. If the service is not behind a proxy or it is not configured appropriately then clients can spoof their IP by just sending certain headers.

The decision to use/trust those headers can only be made by service operators once they know how it will be deployed.

Thanks again for your contribution!

PR #61, opened 2019-02-19
Code to allow shutdown on integrity issue

ardan-bkennedy:


nstogner:
I find it a bit odd that my web framework is sending `os.Signal`s.

ardan-bkennedy:
Hi Nick, there is a special reason for this. We identified a serious integrity issue and the service MUST be shutdown. We should never get to that line of code. We need to make sure the shutdown is graceful as well. This mechanic gives us the ability to do this.

PR #62, opened 2019-02-19
code to shutdown service if context value is missing

ardan-bkennedy:


internal/platform/web/web.go
```diff
@@ -107,3 +108,28 @@ func (a *App) Handle(verb, path string, handler Handler, mw ...Middleware) {
 	// Add this handler for the specified verb and route.
 	a.TreeMux.Handle(verb, path, h)
 }
+
+// Shutdown is a type used to help with the graceful termination of the service.
```
johndopazo: Should this start lowercase `shutdown is ...`.

internal/platform/web/validation.go
```diff
@@ -32,6 +32,12 @@ func (err InvalidError) Error() string {
 	return str
 }
 
+// JSONError is the response for errors that occur within the API.
```
johndopazo: I think moving it to `validation.go` makes it seem like JSONErrors are all because of validation. Is that true?

internal/platform/web/web.go
ardan-bkennedy: Brilliant catch

internal/platform/web/validation.go
ardan-bkennedy: They are because of the `Fields InvalidError` field. I struggled with this but that field sealed the deal for me. Probably the name of the type needs to change.

internal/mid/errors.go
```diff
@@ -19,7 +19,11 @@ func ErrorHandler(before web.Handler) web.Handler {
 		ctx, span := trace.StartSpan(ctx, "internal.mid.ErrorHandler")
 		defer span.End()
 
-		v := ctx.Value(web.KeyValues).(*web.Values)
+		// This will cause the service to be shutdown.
```
jcbwlkr: This comment is confusing. It sounds like that will always do the shutdown but it only does if the ctx value is missing.

internal/mid/errors.go
```diff
@@ -57,6 +60,12 @@ func ErrorHandler(before web.Handler) web.Handler {
 			// Respond with the error.
 			web.Error(ctx, log, w, err)
```
jcbwlkr: Do we care that the shutdown error will display the message `web value missing from context` to the end user?

internal/mid/logger.go
```diff
@@ -21,7 +21,12 @@ func RequestLogger(before web.Handler) web.Handler {
 
 		err := before(ctx, log, w, r, params)
 
-		v := ctx.Value(web.KeyValues).(*web.Values)
+		// This will cause the service to be shutdown.
```
jcbwlkr: Same comment as above.

internal/mid/metrics.go
```diff
@@ -40,9 +40,14 @@ func Metrics(before web.Handler) web.Handler {
 			m.gr.Set(int64(runtime.NumGoroutine()))
 		}
 
+		// This will cause the service to be shutdown.
```
jcbwlkr: Same comment as above.

internal/mid/errors.go
ardan-bkennedy: I was thinking about that. We should

internal/platform/web/validation.go
ardan-bkennedy: You are right on this one. I am moving it back

PR #63, opened 2019-02-20
missing ending  span at internal.platform.web

superwhykz:
Span is not recording. 

image


ardan-bkennedy:
wow, this was a huge catch. I have missed this from the very beginning.

PR #64, opened 2019-03-02
Unmarshal - return error if input fields not match

farrukhny:
Unmarshal not returning any error if field name incorrect or not available. Go version 1.10 added a new method DisallowUnknownFields which we can use to return error for unknown JSON fields.

### DIFF
--- a/internal/platform/web/validation.go
+++ b/internal/platform/web/validation.go
@@ -35,7 +35,11 @@ func (err InvalidError) Error() string {
 // Unmarshal decodes the input to the struct type and checks the
 // fields to verify the value is in a proper state.
 func Unmarshal(r io.Reader, v interface{}) error {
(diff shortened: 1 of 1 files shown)


internal/platform/web/validation.go
```diff
@@ -35,7 +35,11 @@ func (err InvalidError) Error() string {
 // Unmarshal decodes the input to the struct type and checks the
 // fields to verify the value is in a proper state.
 func Unmarshal(r io.Reader, v interface{}) error {
-	if err := json.NewDecoder(r).Decode(v); err != nil {
+
```
ardan-bkennedy: Please remove this linespace and the linespace above the call to decode.Decode and I will accept the PR.

Thank you for this!

farrukhny: I have removed linespace

PR #68, opened 2019-04-09
Apply middleware so they are executed in the order they are listed

jcbwlkr:
Ranging over middleware from beginning to end when building handler chains has the effect of *wrapping* them in the order they are listed but it means the resulting middleware handlers are executed in *reverse* of the order they are shown in the code. This always confused me especially when we have middleware that must be ran after another such as the auth `HasRole` mw which must be ran after the `Authenticate` mw. In the route listing we have to put the `HasRole` first to ensure it is executed after `Authenticate`.

I changed `wrapMiddleware` to loop backwards through the list and wrap in that order. This means when a request comes in then the middleware handlers will be first encountered in the order they are listed in source code.

Also as a minor change I swapped the arguments of `wrapMiddleware`. It was

```
wrapMiddleware(h Handler, mw []Middleware) Handler
```
and is now
```
wrapMiddleware(mw []Middleware, h Handler) Handler
```

To me this reads as "please wrap these middleware around this handler"

Second minor change is the line in `web.go` where we invoke `wrapMiddleware`. I broke it out to two separate lines to make it easier to follow.

### DIFF
--- a/cmd/sales-api/handlers/routes.go
+++ b/cmd/sales-api/handlers/routes.go
@@ -13,7 +13,7 @@ import (
 
 // API returns a handler for a set of routes.
 func API(shutdown chan os.Signal, log *log.Logger, masterDB *db.DB, authenticator *auth.Authenticator) http.Handler {
-	app := web.New(shutdown, log, mid.ErrorHandler, mid.Metrics, mid.RequestLogger)
+	app := web.New(shutdown, log, mid.RequestLogger, mid.Metrics, mid.ErrorHandler)
 
 	// authmw is used for authentication/authorization middleware.
 	authmw := mid.Auth{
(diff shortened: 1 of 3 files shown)


ardan-bkennedy:
I did not experience that when I made that other change. Then the order I listed occurred as shown. This is important. Listing them backwards is so very confusing.

ardan-bkennedy:
I rather list them in the code in the order from first to last and change the internals to deal with that.

jcbwlkr:
In both the `master` branch and this branch, when a request comes in the following happens:

```
mid.RequestLogger   starts
mid.Metrics         starts
mid.ErrorHandler    starts
mid.Authenticate    starts
mid.HasRole         starts
productsList        starts

productsList        returns
mid.HasRole         returns
mid.Authenticate    returns
mid.ErrorHandler    returns
mid.Metrics         returns
mid.RequestLogger   returns
```

In order to achieve that goal with the `master` branch you have to write the code with middleware listed in "**returning order**" like this

```
app := web.New(shutdown, log, mid.ErrorHandler, mid.Metrics, mid.RequestLogger)
app.Handle("GET", "/v1/products", productsList, mid.HasRole, mid.Authenticate)
```

That code doesn't compile but it illustrates my point. Notice the mental leap you have to make to visualize then the `HasRole` MW will run.

The change I am proposing is that the source code should list the middleware in "**starts order**"

```
app := web.New(shutdown, log, mid.RequestLogger, mid.Metrics, mid.ErrorHandler)
app.Handle("GET", "/v1/products", productsList, mid.Authenticate, mid.HasRole)
```

Now if you read top-to-bottom and left-to-right you see the middleware in the order they will start.

PR #70, opened 2019-04-11
Use ochttp to create initial span

jcbwlkr:
Using the `ochttp.Handler` as the initial web application handler deals with creating the initial span with or without a remote parent.

It names the initial span after the request URL which is a good idea IMO. This screenshot is an example of a trace for a login attempt

Screen Shot 2019-04-11 at 2 14 28 PM

It also captures the status code of the response as well as some other metadata which is annotated on the trace. See second screenshot from the same trace:

Screen Shot 2019-04-11 at 2 15 28 PM

### DIFF
--- a/internal/platform/web/web.go
+++ b/internal/platform/web/web.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/dimfeld/httptreemux"
+	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/plugin/ochttp/propagation/tracecontext"
 	"go.opencensus.io/trace"
 )
@@ -36,6 +37,7 @@ type Handler func(ctx context.Context, log *log.Logger, w http.ResponseWriter, r
 // data/logic on this App struct
 type App struct {
 	*httptreemux.TreeMux
+	och      *ochttp.Handler
 	shutdown chan os.Signal
 	log      *log.Logger
 	mw       []Middleware
@@ -43,12 +45,25 @@ type App struct {
 
 // New creates an App value that handle a set of routes for the application.
 func New(shutdown chan os.Signal, log *log.Logger, mw ...Middleware) *App {
-	return &App{
+	app := App{
 		TreeMux:  httptreemux.New(),
 		shutdown: shutdown,
 		log:      log,
 		mw:       mw,
 	}
+
+	// Create an OpenCensus HTTP Handler which wraps the router. This will start
+	// the initial span and annotate it with information about the request/response.
+	//
+	// This is configured to use the W3C TraceContext standard to set the remote
+	// parent if an client request includes the appropriate headers.
+	// https://w3c.github.io/trace-context/
+	app.och = &ochttp.Handler{
+		Handler:     app.TreeMux,
+		Propagation: &tracecontext.HTTPFormat{},
+	}
+
+	return &app
 }
 
 // SignalShutdown is used to gracefully shutdown the app when an integrity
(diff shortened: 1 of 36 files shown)


internal/platform/web/web.go
```diff
@@ -95,11 +96,6 @@ func (a *App) Handle(verb, path string, handler Handler, mw ...Middleware) {
 		}
 		ctx = context.WithValue(ctx, KeyValues, &v)
 
-		// Set the parent span on the outgoing requests before any other header to
-		// ensure that the trace is ALWAYS added to the request regardless of
-		// any error occuring or not.
-		hf.SpanContextToRequest(span.SpanContext(), r)
```
jcbwlkr: Right here calling `SpanContextToRequest` adds the `Trace-Parent` header to the request before it travels down the handler chain. I can't see that we need to keep that.

From what I can tell since we have already started the span with the remote parent that should no longer matter to us since we are a server.

HOWEVER if our server were to make further HTTP requests to other upstream servers then those outgoing requests should carry on the trace info. Our application is not doing that and even if it was I'm not sure if that is how we would want to propagate it down. I guess I'm saying I need convinced that these particular lines are still necessary

ardan-bkennedy: Please leave this. If we decide to make a call like that, we don't want to lose it. We should teach this. It is the correct thing to do.

jcbwlkr: My gut said this code was wrong so I dug into it more and now I can **definitely say** that this code has to go.

The line `hf.SpanContextToRequest(span.SpanContext(), r)` should only be called on outgoing requests and it must be called **as late as possible,** right before the outgoing request is sent to an upstream server. The reason is that outgoing request is tagged not only with the TraceID but also the SpanID of the specific action that triggered the upstream request.

I have a full example here https://gist.github.com/jcbwlkr/e6d7dee6eb614a5f1dbc7eb263adb819

These are the relevant bits

```go
func handler(w http.ResponseWriter, r *http.Request) {
	ctx, span := trace.StartSpan(r.Context(), r.URL.Path)
	defer span.End()

	// Make 2 requests to upstream servers.
	_, _ = fetch(ctx, "gophers")
	_, _ = fetch(ctx, "concurrency")
}

func fetch(ctx context.Context, page string) ([]byte, error) {
	ctx, span := trace.StartSpan(ctx, "fetch/"+page)
	defer span.End()

	// Make a request to send to an upstream server. Annotate it with the span context.
	r, _ := http.NewRequest("GET", "google.com/"+page, nil)
	hf.SpanContextToRequest(span.SpanContext(), r)

        // ...
}
```

These generate a trace like this

```
TraceID: 1e9a728d020d9336ce210cc1f6375fe9
┃
┗━ Span: "/search",  ID: 372ffa5ee2b5aa05
   ┃
   ┗━┳━ Span: "fetch/gophers", ID: 1a542139a37bbda6
     ┃  ┃
     ┃  ┗━ HTTP Request for page: "gophers"
     ┃     GET google.com/gophers HTTP/1.1
     ┃     Traceparent: 00-1e9a728d020d9336ce210cc1f6375fe9-1a542139a37bbda6-01
     ┃
     ┗━ Span: "fetch/concurrency", ID: fd7848136441d047
        ┃
        ┗━ HTTP Request for page: "concurrency"
           GET google.com/concurrency HTTP/1.1
           Traceparent: 00-1e9a728d020d9336ce210cc1f6375fe9-fd7848136441d047-01
```

Notice the headers on the outgoing requests include the SpanID for Span "fetch/gophers" or Span "fetch/concurrency" as appropriate.

So, removing these lines in this PR doesn't lose any information. The span context is already there in the spans we create as we move down the call stack.

We should be teaching this but putting it in `internal/platform/web` is not the place to do it. To teach this we should set up a second server configured to export traces to zipkin and have `sales-api` make a request to that second server.

PR #74, opened 2019-04-22
Extract http client factory into platform/web package and fix typos.

ecampolo:
Avoid code duplication when creating http client by extracting into a factory method under platform/web package.
Fix DualStack deprecated property.
Fix typos.

### DIFF
--- a/internal/platform/db/db.go
+++ b/internal/platform/db/db.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
-	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2"
 )
 
 // ErrInvalidDBProvided is returned in the event that an uninitialized db is
--- a/cmd/sales-api/main.go
+++ b/cmd/sales-api/main.go
@@ -19,7 +19,7 @@ import (
 	"github.com/ardanlabs/service/internal/platform/db"
 	"github.com/ardanlabs/service/internal/platform/flag"
 	itrace "github.com/ardanlabs/service/internal/platform/trace"
-	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/kelseyhightower/envconfig"
 	"go.opencensus.io/trace"
 )
--- a/cmd/sidecar/metrics/main.go
+++ b/cmd/sidecar/metrics/main.go
@@ -79,7 +79,7 @@ func main() {
 	// =========================================================================
 	// Start collectors and publishers
 
-	// Initalize to allow for the collection of metrics.
+	// Initialize to allow for the collection of metrics.
 	collector, err := collector.New(cfg.Collect.From)
 	if err != nil {
 		log.Fatalf("main : Starting collector : %v", err)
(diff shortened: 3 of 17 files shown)


internal/platform/db/db.go
```diff
@@ -81,7 +81,7 @@ func (db *DB) Copy() *DB {
 
 // Execute is used to execute MongoDB commands.
 func (db *DB) Execute(ctx context.Context, collName string, f func(*mgo.Collection) error) error {
-	ctx, span := trace.StartSpan(ctx, "platform.DB.Execute")
+	_, span := trace.StartSpan(ctx, "platform.DB.Execute")
```
ardan-bkennedy: Not sure I agree with this change

ardan-bkennedy:
Thanks for the PR!

I'm good with everything but the use of the blank identifier.

internal/platform/db/db.go
ecampolo: Yes, you have a point here. I will rollback this change shortly.

ecampolo:
Done! Thank you for open sourcing this !

cmd/sidecar/tracer/handlers/zipkin.go
```diff
@@ -24,31 +24,16 @@ type Zipkin struct {
...
-	}
-
 	z := Zipkin{
 		zipkinHost:  zipkinHost,
 		localHost:   localHost,
 		sendTimeout: sendTimeout,
-		client: http.Client{
-			Transport: &tr,
-		},
+		client:      web.NewHTTPClient(),
```
ardan-bkennedy: I don't see where this function is being implemented under web?

ecampolo: Yes, platform/web/client.go

internal/platform/web/client.go
```diff
@@ -0,0 +1,28 @@
...
+	"net/http"
+	"time"
+)
+
+const (
+	// DefaultTimeout is the default amount of time for a request to complete before it times out.
+	DefaultTimeout = 30 * time.Second
+)
+
+func NewHTTPClient() *http.Client {
```
ardan-bkennedy: I don't like this for several reasons.

This is hiding cost for the sake of not repeating yourself. I look at the call and I don't know what the settings are. This is important. This function is not creating a new semantic where things are precise.

Also, do not construct values to variables using pointer semantics. The dialer should be a value and the & should be applied to line 23. The same on the construction of the client.

I would rather take this simplier form of construction the dialer and client and replace this code with the old code.

Don't feel that repeating yourself is wrong. This is code that comes from the standard library and it doesn't need to be abstracted. Plus, if one of the places this is used needs special settings, we lost that. Too much being hidden for no gain.

ardan-bkennedy: Losing the setting is not good either.

https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

Let's not do this at all.

ecampolo: Ok, I try to extract to a factory method a initial implementation just for the sake of DRY but taking into account future customizations like:

https://github.com/tsenart/vegeta/blob/8c061dc2715c10b49d5f47fbaf07c3f4c3bce4c7/lib/attack.go

https://github.com/emacampolo/gomparator/blob/master/internal/platform/http/client.go

https://github.com/hashicorp/go-retryablehttp/blob/master/client.go

ecampolo: I will rollback that change as well

ardan-bkennedy: If we had special platform or busines logic, this would make sense. We are not. This is just basic transport and client construction. It doesn't need the level of abstraction. Just work with these types directly.

jcbwlkr:
Hi, @ecampolo. Thanks for your contribution! I have merged it into master. There were a few things I want to change such as preserving the import aliases but I figured it was easiest to just merge it and change those couple little things.

Thanks again!

PR #77, opened 2019-05-10
fixing idioms and mental models

ardan-bkennedy:
It still has been bothering me that there is no consistency in how we are handling errors, my mental models are off and bolierplate code was missing from the web package.

In not being able to sleep, I think I figured it out. Along the way, found another naming issue.

### DIFF
--- a/cmd/sales-api/handlers/product.go
+++ b/cmd/sales-api/handlers/product.go
@@ -46,9 +46,9 @@ func (p *Product) Retrieve(ctx context.Context, w http.ResponseWriter, r *http.R
 	if err != nil {
 		switch err {
 		case product.ErrInvalidID:
-			return web.RespondError(err, http.StatusBadRequest)
+			return web.ErrRequestFailed(err, http.StatusBadRequest)
 		case product.ErrNotFound:
-			return web.RespondError(err, http.StatusNotFound)
+			return web.ErrRequestFailed(err, http.StatusNotFound)
 		default:
 			return errors.Wrapf(err, "ID: %s", params["id"])
 		}
@@ -67,7 +67,7 @@ func (p *Product) Create(ctx context.Context, w http.ResponseWriter, r *http.Req
 
 	v, ok := ctx.Value(web.KeyValues).(*web.Values)
 	if !ok {
-		return web.Shutdown("web value missing from context")
+		return web.ErrShutdown("web value missing from context")
 	}
 
 	var np product.NewProduct
@@ -93,7 +93,7 @@ func (p *Product) Update(ctx context.Context, w http.ResponseWriter, r *http.Req
 
 	v, ok := ctx.Value(web.KeyValues).(*web.Values)
 	if !ok {
-		return web.Shutdown("web value missing from context")
+		return web.ErrShutdown("web value missing from context")
 	}
 
 	var up product.UpdateProduct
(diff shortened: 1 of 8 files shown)


jcbwlkr:
This change is good. I still think we could do better on the names.

If I see an identifier starting with `Err` like `web.ErrShutdown` I am going to first think it is a sentinel error like `product.ErrNotFound` or `sql.ErrNoRows`. So the fact that `web.ErrRequestFailed` and `web.ErrShutdown` are actually functions which return errors of some type is confusing.

These functions are factory functions so what about this?

 ```
package web

type AppError{
  Err    error
  Status int
  Fields []FieldError
}
func NewAppError(error, int) error


type ShutdownError{
  Message string
}
func NewShutdownError(string) error
func IsShutdown(error) bool
```

Another alternative would be to unexport these types and make the functions named `web.AppError` and `web.ShutdownError`. Since the error response behavior is back in the `web` package, no other packages need to know the concrete types outside of where we use `IsShutdown`.

**Comment by ardan-bkennedy** [comment]
I like the `Err` naming convention for returning error values. I know these are not variables but I figured it wouild convey that same message of return an error value. If you look at the handler code that is using it, the readability I think will be better using `Err` vs `New`.

```
prod, err := product.Retrieve(ctx, dbConn, params["id"])
	if err != nil {
		switch err {
		case product.ErrInvalidID:
			return web.ErrRequestFailed(err, http.StatusBadRequest)
		case product.ErrNotFound:
			return web.ErrRequestFailed(err, http.StatusNotFound)
		default:
			return errors.Wrapf(err, "ID: %s", params["id"])
		}
	}
```
vs
```
prod, err := product.Retrieve(ctx, dbConn, params["id"])
	if err != nil {
		switch err {
		case product.ErrInvalidID:
			return web.NewRequestFailed(err, http.StatusBadRequest)
		case product.ErrNotFound:
			return web.NewRequestFailed(err, http.StatusNotFound)
		default:
			return errors.Wrapf(err, "ID: %s", params["id"])
		}
	}
```
This is complicated because I think we have something new here. Using `New` on the return I think is wrong. Let's focus this decision on what makes this code example more readable.

**Comment by jcbwlkr** [comment]
I agree that `NewRequestFailed` is hard to read but that's not what I suggested. I suggested `return web.NewAppError`. Which is true! We are returning a new error.

Really though it's  "Wrapped" error. I would feel better about something like `return web.WrapError`

On the Shutdown side of things it absolutely is a new error. You give it a string and it gives you a shutdown error. That one definitely should be `return web.NewShutdownError`

**Comment by ardan-bkennedy** [comment]
```
prod, err := product.Retrieve(ctx, dbConn, params["id"])
	if err != nil {
		switch err {
		case product.ErrInvalidID:
			return web.ErrRequestFailed(err, http.StatusBadRequest)
		case product.ErrNotFound:
			return web.NewRequestFailed(err, http.StatusNotFound)
		default:
			return web.WrapRequestError(err, http.StatusWhatever)
		}
	}
```

Reading this code, which is best describing the business behavior? Forget about which one is more technically correct.

**Comment by jcbwlkr** [comment]
if we are returning an error for this request then, in MY mind the most clear thing hasn't been mentioned yet which is `return web.NewRequestError(err, 404)`. 

```go
package web

type RequestError struct{}
func NewRequestError(error, int) error
``` 

Try looking at it my way

```go
prod, err := product.Retrieve(ctx, dbConn, params["id"])
if err != nil {
	switch err {
	case product.ErrInvalidID:
		return web.NewRequestError(err, http.StatusBadRequest)
	case product.ErrNotFound:
		return web.NewRequestError(err, http.StatusNotFound)
	case product.ErrForbidden:
		return web.NewRequestError(err, http.StatusForbidden)
	default:
		return errors.Wrapf(err, "ID: %s", params["id"])
	}
}
```

ardan-bkennedy:
Ok, let’s use New. I’ll make the change later.

ardan-bkennedy:
@jcbwlkr you were right. Your suggestion works well and is more idiomatic.

jcbwlkr:
Looks great. Thanks, man!

PR #78, opened 2019-05-10
Switch from Mongo to Postgres

jcbwlkr:
This PR is a work in progress. Do not merge yet. So far I have

1. Switched the `product` package to use postgres including that package's tests.
2. Added "migrate" and "seed" commands to the `sales-admin` tool.
3. Added `database` and supporting packages.
4. Wired db into `sale-api` main func

I still need to

- [x] Fix integration tests
- [x] Switch User and Health Check
- [x] Remove unused packages
- [ ] Deal with deployment? K8s?


cmd/sales-admin/main.go
```diff
@@ -34,11 +36,8 @@ func main() {
 	// Configuration
 
 	var cfg struct {
-		CMD string `envconfig:"CMD"`
-		DB  struct {
-			DialTimeout time.Duration `default:"5s" envconfig:"DIAL_TIMEOUT"`
-			Host        string        `default:"localhost:27017/gotraining" envconfig:"HOST"`
-		}
+		CMD  string `envconfig:"CMD"`
+		DB   database.Config
```
ardan-bkennedy: Not sure I like this at all. I rather see it and the tags. This is hiding information. Let's not couple these things together.

cmd/sales-admin/main.go
```diff
@@ -64,9 +63,13 @@ func main() {
 	case "keygen":
 		err = keygen(cfg.Auth.PrivateKeyFile)
 	case "useradd":
-		err = useradd(cfg.DB.Host, cfg.DB.DialTimeout, cfg.User.Email, cfg.User.Password)
+		err = useradd(cfg.DB, cfg.User.Email, cfg.User.Password)
```
ardan-bkennedy: I don't remember this switch at all. What is this here for?

cmd/sales-admin/main.go
```diff
@@ -135,3 +139,23 @@ func useradd(dbHost string, dbTimeout time.Duration, email, pass string) error {
 	fmt.Printf("User created with id: %v\n", usr.ID.Hex())
 	return nil
 }
+
+func migrate(cfg database.Config) error {
```
ardan-bkennedy: What is the purpose and do these belong in main?

cmd/sales-api/handlers/product.go
```diff
@@ -4,29 +4,27 @@ import (
...
 	"github.com/ardanlabs/service/internal/product"
+	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 )
 
 // Product represents the Product API method handler set.
 type Product struct {
-	MasterDB *db.DB
+	db *sqlx.DB
```
ardan-bkennedy: curious why this is unexported. I am sure I'll see it soon.

cmd/sales-api/main.go
```diff
@@ -58,10 +58,7 @@ func main() {
 			WriteTimeout    time.Duration `default:"5s" envconfig:"WRITE_TIMEOUT"`
 			ShutdownTimeout time.Duration `default:"5s" envconfig:"SHUTDOWN_TIMEOUT"`
 		}
-		DB struct {
-			DialTimeout time.Duration `default:"5s" envconfig:"DIAL_TIMEOUT"`
-			Host        string        `default:"mongo:27017/gotraining" envconfig:"HOST"`
-		}
+		DB    database.Config
```
ardan-bkennedy: Not sure I like this at all. I rather see it and the tags. This is hiding information. Let's not couple these things together.

docker-compose.yaml
```diff
@@ -20,6 +20,15 @@ services:
...
     command: --bind_ip 0.0.0.0
 
+  # This starts a local PostgreSQL DB.
+  db:
+    container_name: sales_db
+    networks:
+      - shared-network
+    image: postgres:11.1-alpine
+    ports:
+      - 5432:5432
```
ardan-bkennedy: Make sure this binds on 0.0.0.0 or we will have a problem with compose.

internal/platform/database/database.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"net/url"
+
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq" // The database driver in use.
+)
+
+// Config is the required properties to use the database.
+type Config struct {
```
ardan-bkennedy: NO! These tags are hidden. Please remove this and in the cfg struct ask for everything that is needed. Decouple this stuff.

internal/platform/database/database.go
```diff
@@ -0,0 +1,44 @@
...
+func Open(cfg Config) (*sqlx.DB, error) {
+
+	// Define SSL mode.
+	sslMode := "require"
+	if cfg.DisableTLS {
+		sslMode = "disable"
+	}
+
+	// Query parameters.
+	q := url.Values{}
```
ardan-bkennedy: Should this be `var`

internal/platform/database/databasetest/docker.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// container tracks information about a docker container started for tests.
+type container struct {
+	ID   string
+	Host string // IP:Port
+}
+
+// startContainer runs a postgres container to execute commands.
+func startContainer(t *testing.T) *container {
```
ardan-bkennedy: Wonder if we don't pass in the container name as a param?

internal/platform/database/schema/migrate.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/GuiaBolso/darwin"
+	"github.com/jmoiron/sqlx"
+)
+
+// Migrate attempts to bring the schema for db up to date with the migrations
+// defined in this package.
+func Migrate(db *sqlx.DB) error {
+
```
ardan-bkennedy: remove line space

internal/platform/database/databasetest/docker.go
jcbwlkr: No because the function is hard coded to extract postgres specific information from the container once it is running. We *could* let that function take an image name but it has to be some version of postgres.

This is not a generic "start any kind of container" function.

docker-compose.yaml
jcbwlkr: Mongo needed that but as far as I can tell postgres doesn't.

jcbwlkr:
The majority of this is done and ready for review. I need to run `go mod vendor` but I am waiting until it is reviewed.

This also does not address the k8s stuff.

andreygrehov:
I would rather introduce a repository DDD pattern and provide an implementation for Postgres.

ardan-bkennedy:
@andreygrehov I would need to hear more about this. We do have a very specific design pattern and a set of guidlines we are following. If you would like to jump on a video chat to discuss your ideas, I would like that.

PR #89, opened 2019-06-12
CI: Use machine executor.

felicianotech:
I made a comment about Go. Otherwise this is a start to changing the config over.


jcbwlkr:
Thanks!

PR #92, opened 2019-06-20
adding the logging of Go's stack trace

ardan-bkennedy:
We are not logging Go's stack trace for the panic. This is very important.


jcbwlkr:
Ah, right. So I considered this when I split the panic middleware out. We are already capturing a stack trace in this middleware when we call errors.Errorf. I was counting on that stack trace being printed when the error is handled in the Errors middleware. And at the time it was. Is it not doing that anymore?

ardan-bkennedy:
More important it’s not the same stack trace. What the errors package does is independent of Go.

PR #105, opened 2019-10-12
added some logic to makefile 

rjtch:
I added some logic in the Makefile to start the webserver with just just one command make all and with make start the project will be started without build and with make up the docker-compose will pull additional information from cloud.google before starting.

### DIFF
--- a/makefile
+++ b/makefile
@@ -54,10 +54,10 @@ metrics:
 		--build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` \
 		.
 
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
I’d rather not have all do up. There are times I just need to build

What does pull do?

rjtch:
i always use the pull to get docker service images before running from a docker trusted registry.  And about getting all the services up i can also provide a command in Makefile to build the images.

ardan-bkennedy:
I made changes to support kind and I need to fix some workflow. I am going to close this PR. Let's talk when you get the chance.

PR #130, opened 2020-06-23
Removed explicit call to os.Exit(1).

sreejeet:
os.Exit(1) is called implicitly by log.Fatalln

Or is there some specific reason for this to be called explicitly, that I am missing?

### DIFF
--- a/app/sales-api/main.go
+++ b/app/sales-api/main.go
@@ -43,8 +43,7 @@ func main() {
 	log := log.New(os.Stdout, "SALES : ", log.LstdFlags|log.Lmicroseconds|log.Lshortfile)
 
 	if err := run(log); err != nil {
-		log.Println("main: error:", err)
-		os.Exit(1)
+		log.Fatalln("main: error:", err)
 	}
 }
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
I don't want to do this. I teach specifically how the call to Fatal is hiding information. Readability is about not hiding the cost of things. I rather be specific and log and call os.Exit(1).

sreejeet:
As a training material, what you say does make sense.
But would my proposal be valid in case of a real world application where the cost of such functions is well understood by the developers?

ardan-bkennedy:
No, this is real world production code (to the extent of still keeping it simple). Obviously things I am doing with keys and tokens can't be done and that is documented.

it is better to make things easy to understand than easy to do.

So in a code review for a "real world" app I would still reject this change. You assume this is well understood and it is not.

sreejeet:
> You assume this is well understood and it is not.
Right, I get your point now. Even I learned about this only after referring to the documentation.

Thanks for clarifying.

PR #133, opened 2020-07-19
updated tracer info

ssenthil416:
Issue : folder name  "internal" and "platform" was changed to "business" and "foundation", but tracer string was not updated. Sample in database.go  "platform.database.statuscheck" 

Fix: So changed  internal to business and platform to foundation. Sample  "foundation.database.statuscheck"


ardan-bkennedy:
Thanks for seeing this!

PR #140, opened 2021-01-16
Fix Trace Log for Business User Data

didiyudha:
Hi Ardan Labs Team,

I think there's a little mistake on package `github.com/ardanlabs/service/business/data/user` , exactly on trace log. Here's the details:

Before fixing: `ctx, span := trace.SpanFromContext(ctx).Tracer().Start(ctx, "internal.data.user.create")`
After fixing: `ctx, span := trace.SpanFromContext(ctx).Tracer().Start(ctx, "business.data.user.create")`

Thank you.


ardan-bkennedy:
Sure is. Thanks!!

PR #144, opened 2021-02-05
fix empty slice declaration via literal

Le0nX:


ardan-bkennedy:
I want the empty slice construction here since there is a change a query might return an empty result. The construction helps to document that.

Le0nX:
@ardan-bkennedy ok, got it. Thx for your training!

PR #151, opened 2021-03-13
fixed publicFile filename in keygen.go

zue666:
should be `publicFile` instead of `privateFile`.


ardan-bkennedy:
Nice catch!!! Can you put your name/email into the contributors file please.

PR #154, opened 2021-03-21
fix - use shutdown error in authorize when claims are missing

kinluek:
Forgive me if I'm wrong but in the context of your application I think missing claims in the authorize middleware should return a shutdown error as this should only ever be called if we have been successfully authenticated first?

### DIFF
--- a/business/mid/auth.go
+++ b/business/mid/auth.go
@@ -68,7 +68,7 @@ func Authorize(roles ...string) web.Middleware {
 			// If the context is missing this value return failure.
 			claims, ok := ctx.Value(auth.Key).(auth.Claims)
 			if !ok {
-				return errors.New("claims missing from context")
+				return web.NewShutdownError("claims missing from context")
 			}
 
 			if !claims.Authorized(roles...) {


ciprian-cimpan:
In the `Logger` middleware, the service is requesting a shutdown when claims are missing, so this would also be better for consistency.

kinluek:
> In the `Logger` middleware, the service is requesting a shutdown when claims are missing, so this would also be better for consistency.

@ciprian-cimpan - The logger middleware is actually checking for trace values to correlate the logs and time requests, which is a bit of different concern to the problem I've pointed out... not really a consistency issue I don't think

ardan-bkennedy:
It’s not necessary to shutdown here. Since not having claims means you can’t execute any logic, nothing can be in a lack of integrity. 

The other handlers signal a shutdown because if it’s missing, something very bad is going on. This is not the case with auth.

ardan-bkennedy:
Notice what we are looking for in the context is also different. It’s generated by middleware, not the framework. It’s not necessarily expected to be there. 

If you agree then please close the PR.

kinluek:
I do see the difference but the **Authorize** middleware should only be used in conjunction with the **Authenticate** middleware and should only be called once the **Authenticate** middleware has authenticated the user, as stated in the function comments - `"Authorize validates that an **authenticated** user..."`. Therefore if the Authorize middleware is missing the claims this would mean that someone has constructed the application incorrectly, in this instance would we not want to terminate the program and add a fix? 

Is there any instance where the program would be running as intended or correctly and managing to find no claims in the Authorize middleware?

ardan-bkennedy:
Let me think about that more. It’s interesting if you are only talking about the authorize middleware. 

Give me till Monday.

kinluek:
Sure thing, and yes this is only for the authorize middleware

ardan-bkennedy:
I still believe a shutdown is not warranted here. A shutdown is serious and needs to be used when we have an integrity issues that would affect all the requests. I don't think this is the case in this situation.

Even if someone forgot to add the Authenticate middleware but added Authorize, those requests would not be allowed to get to the handler. So that alone is stopping an integrity issue. Even though the middleware is not applied properly, this doesn't affect every request.

My guess is, local handler middleware shouldn't cause a shutdown in general. Application level middleware has that option.

kinluek:
Yeh okay, makes sense, maybe it just needs a more helpful error message?

ardan-bkennedy:
Excellent point. Will be publishing a change for that soon.

PR #156, opened 2021-03-21
Allow testing against multiple database versions

sudomateo:
I've been following this repository in Ultimate Service v1.0 and Ultimate Service v2.0. Thank you for making this code available to everyone.

I have a similar need to spin up a database container to run tests against. However, as I was reading the code in this repository I asked myself "How can I test against multiple database versions?". These changes allow for that by using Go sub tests. I'm not entirely sure if this is the best implementation, but I'm open for discussion!

### DIFF
--- a/app/sales-api/tests/product_test.go
+++ b/app/sales-api/tests/product_test.go
@@ -32,22 +32,26 @@ type ProductTests struct {
 // subtest needs a fresh instance of the application it can make it or it
 // should be its own Test* function.
 func TestProducts(t *testing.T) {
-	test := tests.NewIntegration(t)
-	t.Cleanup(test.Teardown)
+	for name, tc := range tests.Databases {
+		t.Run(name, func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


ardan-bkennedy:
At first glance things look pretty good. On Monday I will take a closer look. I have to also consider if this functionality belongs in this project or if this is specific to your use case.

I think we can add the support but I wouldn’t have the tests actually run against two different database versions.

Let’s focus on the tests package to start. Then have the tests still just run again a single version. For your use case, the support is there.

ardan-bkennedy:
Applied a change to supporty this based on some of your changes. The NewUnit and NewIntegration functions now take a DBContainer value. This will allow projects to write table tests against different db versions.

sudomateo:
Thanks Bill!

ardan-bkennedy:
Thanks for the suggestion and taking the time to provide a solution.

Issue #157, opened 2021-03-23
CORS & SetMaxOpenConns


jonleopard:
Hello!

I'm going over this course again and noticed that CORS wasn't really addressed in any of the lectures. Do you think it might be valuable to add a small section for that in the course? Also, `SetMaxOpenConns` is not being defined in the config.  I read this very interesting [article](https://sudhir.io/understanding-connections-pools/) that suggests it being set. Here's an excerpt:

> Because Go is automatically geared for tens of thousands of simultaneous requests even on very small servers, pairing it with a process-based database like PosgreSQL is often like driving a race-car into a brick wall. If each goroutine is using the standard library SQL package, it will create as many connections as there are goroutines—because the default pool size is unlimited. The first thing you'd want to do on any Go application using a SQL DB is to configure the connection limits using `SetMaxOpenConns` and the associated options. Go uses an internal pool on the sql package, so each query you run will checkout a connection from the pool, use it and release it back immediately. This means that if you want to execute a transaction, you must use the special methods that give you an object that wraps the connection you started the transaction on, which is the only way you can later commit or rollback that transaction. This is also every important when using other functions that are connection specific, like prepared statements or advisory locks.

Let me know what you think!

Cheers :)

ardan-bkennedy:
Hi Jon, I will leave this issue here and look at this ASAP. I have a few other pending things to get to.

I've implemented CORS before but it was a while ago. I would like to consult a friend of mine on this. Just need to think of a proper use case.

I'll read the article as well on the open conns.

Thanks!!

ardan-bkennedy:
I think this should be applied at the route level unless every single route needed these headers.

I like the solution.

Create a PR adding this middle to the mid package and the new config setting with that default.

I won't apply this to any route at this time.

Make sure you add yourself to the CONTRIBUTORS files as well please.

jonleopard:
@ciprian-cimpan Thanks for the info! I'll go ahead and wait for the new PR to be merged and pull it in whenever its finished.

ciprian-cimpan:
@jonleopard you're welcome! 👍🏼 
I just submitted the [PR](https://github.com/ardanlabs/service/pull/160)

ardan-bkennedy:
@jonleopard I added the Idle/Open connection settings.

PR #160, opened 2021-03-24
Added middleware for CORS support and the config option needed by it

ciprian-cimpan:
As per [Issue 157](https://github.com/ardanlabs/service/issues/157):
1 - A CORS middleware has been added to the `business/web/mid` layer  
2 - The option for the allowed origin support has been added to the `sales-api` main config
3 - Support for CORS `OPTION` preflight requests was added to the `sales-api` handlers

Regarding the 3rd item on the list, this was a quick hack in my case and I'm not entirely sure this is the most elegant way to do it.
Although I like it being bundled with all the other available routes in the handlers file, I think that ideally this should be hidden "under the hood", maybe around `foundation/web -> handle()`? 
I think a case could be made that this should be part of foundation, as a web service CORS is quite _foundational_, not entirely sure though...

### DIFF
--- a/app/sales-api/handlers/handlers.go
+++ b/app/sales-api/handlers/handlers.go
@@ -3,6 +3,7 @@
 package handlers
 
 import (
+	"context"
 	"log"
 	"net/http"
 	"os"
@@ -16,10 +17,10 @@ import (
 )
 
 // API constructs an http.Handler with all application routes defined.
-func API(build string, shutdown chan os.Signal, log *log.Logger, a *auth.Auth, db *sqlx.DB) http.Handler {
+func API(build string, shutdown chan os.Signal, log *log.Logger, a *auth.Auth, db *sqlx.DB, corsOrigin string) http.Handler {
 
 	// Construct the web.App which holds all routes as well as common Middleware.
-	app := web.NewApp(shutdown, mid.Logger(log), mid.Errors(log), mid.Metrics(), mid.Panics(log))
+	app := web.NewApp(shutdown, mid.Cors(corsOrigin), mid.Logger(log), mid.Errors(log), mid.Metrics(), mid.Panics(log))
 
 	// Register debug check endpoints.
 	cg := checkGroup{
@@ -28,6 +29,11 @@ func API(build string, shutdown chan os.Signal, log *log.Logger, a *auth.Auth, d
 	}
 	app.HandleDebug(http.MethodGet, "/readiness", cg.readiness)
 	app.HandleDebug(http.MethodGet, "/liveness", cg.liveness)
+
+	// Accept CORS 'OPTIONS' preflight requests
+	app.Handle(http.MethodOptions, "/*", func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+		return nil
+	})
 
 	// Register user management and authentication endpoints.
 	ug := userGroup{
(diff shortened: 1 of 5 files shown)


jonleopard:
I suppose this might be a dumb question, but is there any added benefit to using something like https://github.com/rs/cors here?

ardan-bkennedy:
I think just another dependency that may not be needed to just apply the headers.

app/sales-api/handlers/handlers.go
```diff
@@ -16,10 +17,10 @@ import (
 )
 
 // API constructs an http.Handler with all application routes defined.
-func API(build string, shutdown chan os.Signal, log *log.Logger, a *auth.Auth, db *sqlx.DB) http.Handler {
+func API(build string, shutdown chan os.Signal, log *log.Logger, a *auth.Auth, db *sqlx.DB, corsOrigin string) http.Handler {
 
 	// Construct the web.App which holds all routes as well as common Middleware.
-	app := web.NewApp(shutdown, mid.Logger(log), mid.Errors(log), mid.Metrics(), mid.Panics(log))
+	app := web.NewApp(shutdown, mid.Cors(corsOrigin), mid.Logger(log), mid.Errors(log), mid.Metrics(), mid.Panics(log))
```
ardan-bkennedy, line 23: Let's not apply it right now to the application middleware

app/sales-api/handlers/handlers.go
```diff
@@ -29,6 +30,11 @@ func API(build string, shutdown chan os.Signal, log *log.Logger, a *auth.Auth, d
 	app.HandleDebug(http.MethodGet, "/readiness", cg.readiness)
 	app.HandleDebug(http.MethodGet, "/liveness", cg.liveness)
 
+	// Accept CORS 'OPTIONS' preflight requests
```
ardan-bkennedy, line 33: Can we comment this code with a good set of notes above. Talk about the need to add the middleware and more of what this does, how and why?

business/web/mid/cors.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// This is the actual middleware function to be executed.
+	m := func(handler web.Handler) web.Handler {
+
+		// Create the handler that will be attached in the middleware chain.
+		h := func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+			ctx, span := trace.SpanFromContext(ctx).Tracer().Start(ctx, "business.mid.cors")
+			defer span.End()
+
+			// Set CORS Headers
```
ardan-bkennedy, line 22: // Set the CORS Headers for the response.

ardan-bkennedy:
I made other changes to the project so I will pick this code out and add it myself.  Thanks for the code!!!

ardan-bkennedy:
@jonleopard, look at the project. Added the idea of optional parameters for CORS since not every project needs this. Add some logic to apply the Options handler, but the user will need to set the middleware. This project doesn't need it.

jonleopard:
@ardan-bkennedy Looks great! Just for my own education, a REST API won't need a CORS policy if both the consumer (client) and API are talking to each other internally (meaning within the same network)? When I made the request from the client, I noticed that there was a CORS error in the console so that's sort of what put me on this trail.

Thanks again!

kinluek:
@jonleopard  - only browsers use the CORS headers, it's not a matter of being on the same network, you could communicate from a server on a different network to your API without CORS access

PR #161, opened 2021-03-26
Replaces TrimRight with TrimSuffix

erkattak:
`strings.TrimRight("develop.pem", ".pem")` results in "develo" because any of the matching characters on the right are trimmed


ardan-bkennedy:
Didn't know that. Thanks for the fix.  Can you please add yourself to the CONTRIBUTORS file before I merge this?

PR #171, opened 2021-06-19
Improves makefile and fixes latest Kind v0.11 breaking changes

ioagel:
## Related to Kind section

Latest Kind `v0.11` brings breaking changes as well as support for both amd64 and arm64 in node images.

To update the project to be compatible with the latest Kind (but break compatibility with old versions) we need to use the node images from this [link](https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1). Using the latest node image from `dockerhub` is not compatible with Kind version >= 0.11, that's why having the comment in the makefile ensures that Kind and node version are in sync.

Furthermore, now we do not need the `make kind-up-m1` target, because the latest Kind and node image support arm64 besides amd64. (To be certain before considering merging the pull request someone with m1 mac should test it)

I also extracted the kind cluster name used: `ardan-starter-cluster` to the variable `KIND_CLUSTER` to improve maintenance and readability. 

So, someone testing the pull request should do: e.g. `brew upgrade kind` and everything related to `Kind` section should work properly for both amd64 and arm64 machines.

# Related to Docker section

To preserve the naming conventions used in the `Kind` section , I adjusted the targets to be more precise and clear.

- The target `docker-down` is simpler and faster. Our end goal is to remove `all containers` running and/or stopped so a gracefull shutdown is not needed and we can use one line to force stop (SIGKILL) and remove every container.
- The variable `FILES` is not needed as we now use the shell command in just one place, because...
-  ... Using it in the last target `docker-kind-logs`, results in buggy behavior. `docker logs` command expects only one container, if we use the list `FILES` and we have more than one container then it errors out.
- As I understand, the `docker logs` command is used only for inspecting the kind node container so using as a target `docker-kind-logs` is more precise .

### DIFF
--- a/makefile
+++ b/makefile
@@ -18,7 +18,7 @@ export PROJECT = ardan-starter-kit
 # Used to install expvarmon program for metrics dashboard.
 # go install github.com/divan/expvarmon@latest
 
-# // To generate a private/public key PEM file.
+# To generate a private/public key PEM file.
 # openssl genpkey -algorithm RSA -out private.pem -pkeyopt rsa_keygen_bits:2048
 # openssl rsa -pubout -in private.pem -out public.pem
 # ./sales-admin genkey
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
This looks great. Thank you for taking the time to do this.

PR #173, opened 2021-06-20
NewString() is the equivalent of New().String()

liyongjing:
v1.2.0 release the NewString() function which is the equivalent of uuid.New().String().


ardan-bkennedy:
The NewString change is good but time is already in UTC. We don’t need that change.

liyongjing:
but now.UTC() in Create function, so delete UTC() too?

ardan-bkennedy:
The web framework should be using UTC, so no other code should need to convert it

PR #174, opened 2021-07-01
keep clean main.go, multiplexing init func

liyongjing:
if have many api, maybe multiplexing init func


ardan-bkennedy:
Don’t want to hide this outside of main. This is very important to have as transparent and readable. 

I want to avoid encapsulation anything from main.

liyongjing:
oh, close it

business/config/entry.go
```diff
@@ -0,0 +1,53 @@
+package config
+
+import "go.uber.org/zap"
+
+func RunApi(log *zap.SugaredLogger, build string) error {
```
andreygrehov, line 5: nit: should be RunAPI.

ardan-bkennedy:
Good point. I’ll make that change.

PR #185, opened 2021-09-22
Fix for product core alias typo

ciprian-cimpan:
Import alias `userProduct` was renamed to `productCore`, which better describes the package and follows the pattern in `usergrp`.


ardan-bkennedy:
This is a good change.

PR #188, opened 2021-10-01
[Fix] Add back authorization checks for product update & delete. 

ciprian-cimpan:
Authorization for product `update` and `delete` actions went missing after the recent `core` updates. 

This fix adds both checks back into the handlers layer.


ardan-bkennedy:
Good Catch

Issue #195, opened 2021-10-21
CORS Headers aren't applied to OPTIONS calls


lgarciaaco:
When constructing the handler with CORS support:
```
	// Construct the mux for the API calls.
	apiMux := handlers.APIMux(handlers.APIMuxConfig{
		Shutdown: shutdown,
		Log:      log,
		Auth:     auth,
		DB:       db,
		Broker:   broker,
	}, handlers.WithCORS("*"))
```

All preflights OPTIONS should pass as intended in https://github.com/ardanlabs/service/blob/master/app/services/sales-api/handlers/handlers.go#L63, yet this only accepts the OPTIONS calls but it doesn't set the response headers needed for Cross-Origin Resource Sharing.

ardan-bkennedy:
Good catch. Will close this and contrinue on the PR.

PR #196, opened 2021-10-21
Fix bug where headers weren't applied to OPTIONS call

lgarciaaco:
Fixes https://github.com/ardanlabs/service/issues/195

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -41,6 +41,7 @@ Jeremy Stone <slycrel@gmail.com>
 John Reese <john@reese.dev>
 Leeyong Jing leeyongjing@gmail.com
 liyongjing <liyongjing@users.noreply.github.com>
+Luis Garcia Acosta <lgarciaac@gmail.com>
 Michael Gasch <mgasch@vmware.com>
 Nick Stogner <nstogner@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


ciprian-cimpan:
I was stupid - my app works because CORS is applied globally. So please disregard my last comment.

ciprian-cimpan:
In my case 90% of the routes need it, so I took the shortcut of adding it to the whole app. 
That's what generated the confusion around this PR, as I couldn't understand at first why it was working for me. 🤷😅

Regarding the current implementation, I think it also depends on your main user target & overall philosophy for this service. If you consider it'll have a lot of adoption as a "backend" for web apps, then `cors` becomes a component that gains "foundational" status, in my view. 

In such a case I would move the handle for preflights in `foundation/web`, allow multiple origins, increase the visibility of the cors middleware being applied to routes _(maybe with a default example)_; or even going as far as applying it by default to all routes _(maybe with a toggle setting?)_.

But I can as well imagine lots of use-cases where what you've built is used as a quick scaffolding for a micro-service that doesn't interact at all with browsers, and then carrying around all this extra `cors` "weight" becomes suboptimal.

Thanks for the fix Luis 🙌👏 - I used this PR to refresh my cors knowledge ❤️

PR #204, opened 2021-11-24
kind-shell command should grep for sales

burhan:
The two shell commands `kind-shell` and `gcp-shell` filter for two different keys. The correct version should grep for `sales` and not `app`.


ardan-bkennedy:
Yea I changed this and missed it. I haven’t tried  GCP in a while but this is a good fix, thanks.

PR #205, opened 2021-11-25
feat: taskfile

burhan:
The current `makefile` is Linux and macOS specific. Although it can be run on Windows using WSL2, this PR adds a [`Taskfile.yml`](https://taskfile.dev/) that has the same commands and can be used across multiple operating systems without relying on shell specific utilities (like `cut`, `grep`, etc.)

It requires the installation of the `task` command, which can be done with `go install github.com/go-task/task/v3/cmd/task@latest`, then all instances of `make` can be replaced with `task`.

```
C:\Users\burhan\service>task tidy
task: [tidy] go mod tidy
go: downloading go.uber.org/zap v1.19.1
...
task: [tidy] go mod vendor
```
Upon merging, the wiki should be updated with instructions for Windows.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,0 +1,309 @@
+version: '3'
+
+# ==============================================================================
+# Testing running system
+
+# For testing a simple query on the system. Don't forget to `make seed` first.
+# curl --user "admin@example.com:gophers" http://localhost:3000/v1/users/token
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
Though I appreciate this I’m not going to accept it. I can’t add another tool for working with the project. I get this affects window users but they can translate the commands as you did. 

Thank you for this.

PR #206, opened 2021-11-29
chore: spelling errors

burhan:
I noticed a few typos and spelling errors while reading through the code.

### DIFF
--- a/foundation/keystore/keystore.go
+++ b/foundation/keystore/keystore.go
@@ -16,7 +16,7 @@ import (
 )
 
 // KeyStore represents an in memory store implementation of the
-// KeyStorer interface for use with the auth package.
+// KeyStore interface for use with the auth package.
 type KeyStore struct {
 	mu    sync.RWMutex
 	store map[string]*rsa.PrivateKey
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,7 +2,7 @@
 .git/
 .reading/
 .dockerignore
-.gitgnore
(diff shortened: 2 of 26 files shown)


foundation/keystore/keystore.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 // KeyStore represents an in memory store implementation of the
-// KeyStorer interface for use with the auth package.
+// KeyStore interface for use with the auth package.
```
ardan-bkennedy, line 19: This is wrong. It should stay KeyStorer. I am sure the interface is named that.

ardan-bkennedy:
Fix the one comment and thanks for the PR. Is your name already in the CONTRIBUTORS file? If not, please add it in alphabetical order to the PR as well.

foundation/keystore/keystore.go
bk-tfo: I based this on the module level comment which mentions `auth.KeyStore` interface. Honestly I was not able to find a reference for `Keystorer` interface.

ardan-bkennedy: The interface is called KeyStore. This comment should remain as is.

ardan-bkennedy: oh, i see you fixed it actually.

ardan-bkennedy: This should be called KeyStorer so I will fix that after this PR.

PR #211, opened 2022-01-06
replace depreciated type

Penthious:
# What
Upgrade jwt package to 4.2
Remove depreciated type: https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Parser


ardan-bkennedy:
This is great, thanks!

PR #214, opened 2022-01-15
runs images as non-root user

ioagel:
- Following security best practices, both **sales-api** and **metrics** docker images are refactored to run as non-root user.
- Fixed broken `make kind-shell` by changing the container name to the proper one.

### DIFF
--- a/makefile
+++ b/makefile
@@ -143,7 +143,7 @@ kind-context-sales:
 	kubectl config set-context --current --namespace=sales-system
 
 kind-shell:
-	kubectl exec -it $(shell kubectl get pods | grep sales | cut -c1-26) --container app -- /bin/sh
+	kubectl exec -it $(shell kubectl get pods | grep sales | cut -c1-26) --container sales-api -- /bin/sh
 
 kind-database:
 	# ./admin --db-disable-tls=1 migrate
--- a/zarf/docker/dockerfile.metrics
+++ b/zarf/docker/dockerfile.metrics
@@ -27,8 +27,11 @@ RUN go build -ldflags "-X main.build=${BUILD_REF}"
 FROM alpine:3.15
 ARG BUILD_DATE
 ARG BUILD_REF
-COPY --from=build_metrics /service/app/services/metrics/metrics /service/metrics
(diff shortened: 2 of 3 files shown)


ardan-bkennedy:
Super interesting. Give me a day to grok this. Thanks!!!!

ardan-bkennedy:
Can you please add comments to the new code. This is new to me and I need them. Mostly what you are doing and why.

ioagel:
The command:
```shell
RUN addgroup -g 1000 -S sales && \
    adduser -u 1000 -h /service -G sales -S sales
```
results in the following changes in our container:
- Creates a new system `-S` *user* with it's associated system `-S` *group* both named **sales**, manually setting the id to `-u|-g 1000` for both.
- The new user is assigned `-h /service` as it's home dir matching the one were our service binary and code will exist.
If in the future we need our service to write files to disk, by setting the home dir we will have the appropriate permissions.
- The reasoning for creating a `system` user instead of a normal one, was for semantic correctness as our user is really a service account. Also we do not have to provide a password interactively breaking our `docker build`.
System users(and groups) when the ID is not manually set, are assigned Ids from 999 and below. However those Ids are the available pool for OS daemons and various system services, so to be on the safe side and prevent having conflicts with future development in our base image `alpine` we chose the first available ID (1000) in *userspace* mode.

The commands:
```shell
COPY --from=build_sales-api --chown=sales:sales /service/zarf/keys/. /service/zarf/keys/.
COPY --from=build_sales-api --chown=sales:sales /service/app/tooling/sales-admin/sales-admin /service/sales-admin
COPY --from=build_sales-api --chown=sales:sales /service/app/services/sales-api/sales-api /service/sales-api
```
copy our files to the image and on the fly set the permissions to our newly created user and group. This a recent addition to docker and saves us the step `RUN chown -R sales:sales /service`

Lastly, to state that the container will run our service as our newly created user we need to add `USER sales` just before  `CMD`

ardan-bkennedy:
One last thing, add yourself to the CONTRIBUTORS file please.

ioagel:
I am already in the CONTRIBUTORS file.
Thanks

PR #220, opened 2022-02-09
Add make target to install Mac dev dependencies

JonnieDoe:
Follow up after https://github.com/ardanlabs/euclass/pull/1

cc @ardan-bkennedy

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -28,6 +28,7 @@ Askar Sagyndyk <superwhykz@gmail.com>
 Bob Cao <3308031+bobintornado@users.noreply.github.com>
 Burhan Khalid <burhan.khalid@gmail.com>
 Ciprian Cimpan <hello@ciprian.pro>
+David Paraschivescu <david.cristian.paraschivescu@gmail.com>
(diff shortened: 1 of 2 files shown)


CONTRIBUTORS
```diff
@@ -56,3 +56,4 @@ Wietse Venema <wietsevenema@gmail.com>
 William Kennedy <bill@ardanlabs.com>
 Wyatt Johnson <wyattjoh@gmail.com>
 Zachary Johnson <zachjohnsondev@gmail.com>
+David Paraschivescu <david.cristian.paraschivescu@gmail.com>
```
ardan-bkennedy, line 59: This needs to be in alphabetical order please.

makefile
```diff
@@ -36,6 +36,16 @@ SHELL := /bin/bash
 # Launch zipkin.
 # http://localhost:9411/zipkin/
 
+
+# ==============================================================================
+# Install dependencies
+
+dev.setup.mac:
+	brew update
```
ardan-bkennedy, line 44: I realize now why we need to run update, to make sure we get the latest version of kid and kustomize available. We need to add kubectl to this list as well.

PR #224, opened 2022-03-23
[Fix] Added OptionHandler on web.App & updated CORS documentation.

cip8:
`web.App` is missing an `OptionHandler` - this PR adds it and updates the CORS middleware docs.

### DIFF
--- a/business/web/v1/mid/cors.go
+++ b/business/web/v1/mid/cors.go
@@ -10,9 +10,9 @@ import (
 /*
 	// This handler needs to be set to be fully compliant with CORS. Do this
 	// in the Routes function.
-	app.OptionsHandler = func(w http.ResponseWriter, r *http.Request, params map[string]string) {
+	app.OptionsHandler(func(w http.ResponseWriter, r *http.Request, params map[string]string) {
 		w.WriteHeader(http.StatusOK)
-	}
+	})
 */
 
 // Cors sets the response headers needed for Cross-Origin Resource Sharing
--- a/foundation/web/web.go
+++ b/foundation/web/web.go
@@ -102,3 +102,8 @@ func (a *App) Handle(method string, group string, path string, handler Handler, 
 	}
 	a.mux.Handle(method, finalPath, h)
 }
+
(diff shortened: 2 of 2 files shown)


ardan-bkennedy:
I made some changes, they are not working for my blockchain project. I really don't understand this stuff.

I had to make sure the Options call also returned these headers. I don't believe I was doing that before?

ardan-bkennedy:
I reverted the changes. I had the right solution in the handlers at the app level all along. I am now using this code in the blockchain and it's working.

cip8:
Hmmm... 🤔 this is strange and I'm not sure why it doesn't work in your case. 

In my app everything works well, including your last refactor.
To summarize how I handle this:

1. Inside `APIMux()`, I have only the following code:

~~~
if opts.corsOrigin != "" {
app.OptionsHandlerAll()
}
~~~

Note: the CORS middleware can be applied globally or for each route individually, the code above is only for `OPTIONS` requests.

2. In `web.go`, the two new methods are like this:
~~~
// OptionsHandler assigns the given Handler to respond to Option requests.
func (a *App) OptionsHandler(handler Handler) {
	h := func(w http.ResponseWriter, r *http.Request, params map[string]string) {
		w.Header().Set("Access-Control-Allow-Origin", "http://localhost:5000")
		w.Header().Set("Access-Control-Allow-Methods", "OPTIONS")
		w.Header().Set("Access-Control-Allow-Headers", "Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization")

		handler(r.Context(), w, r)
	}

	a.mux.OptionsHandler = h
}

// OptionsHandlerAll returns a 200 Status OK for all Option requests.
func (a *App) OptionsHandlerAll() {
	a.OptionsHandler(func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
		return nil
	})
}
~~~

Note: the origin domain can be passed as a function param, I just hardcoded it here for testing.

ardan-bkennedy:
For sure I was not sending headers with the option call in the first refactor. I've seen my blockchain Chrome extension work and not work. It's all a mystery to me.

I like the current implementation. It keeps it at the app layer where it belongs.

PR #226, opened 2022-04-09
using pointer semantic on value slice range

theredrad:
The DB product slice uses value semantics. While ranging on it to convert to the core product, we can use pointer semantic instead of value semantic to prevent an extra copy on the range.
I know it's not too much optimization. 😅

### DIFF
--- a/business/core/product/models.go
+++ b/business/core/product/models.go
@@ -49,8 +49,8 @@ func toProduct(dbPrd db.Product) Product {
 
 func toProductSlice(dbPrds []db.Product) []Product {
 	prds := make([]Product, len(dbPrds))
-	for i, dbPrd := range dbPrds {
-		prds[i] = toProduct(dbPrd)
+	for i := range dbPrds {
+		prds[i] = toProduct(dbPrds[i])
 	}
 	return prds
 }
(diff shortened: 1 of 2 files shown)


ardan-bkennedy:
I hear you but unless the profiler is saying this is a performance issues with the service, it's a premature optimization that is taking an exception to maintaining consistent data semantics.

Let's keep with the data semantic consistency until we can't.

theredrad:
I see this engineering decision about consistency and it's beautiful. 
Also, I ran a benchmark test for these semantics:

```
goos: darwin
goarch: amd64
pkg: valuepointer
cpu: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
BenchmarkRangeValue-8               4080            269524 ns/op          647169 B/op          1 allocs/op
BenchmarkRangeValuePointer-8        4957            243476 ns/op          647171 B/op          1 allocs/op
PASS
```

`BenchmarkRangeValue` benchmarks the range for value slice using value semantic (i, val := range)
`BenchmarkRangeValuePointer` benchmarks the range for value slice using pointer semantic (i := range)

It's not too much 😅.

Feels free to close the pull request.

ardan-bkennedy:
Yea, 26 microseconds. This is usually the case with most things internally.

PR #227, opened 2022-06-16
omit duplicate copy in metrics docker file

AhmedShaef:
found this line duplicate
```
 # Copy the source code into the container.
 COPY . /service
```


ardan-bkennedy:
Not sure how that happend!! Thanks!

PR #228, opened 2022-06-27
handle error that occurs when the client disconnects while the http.ResponseWriter is about/being to be written to

jtrinh27:
Gracefully handle write errors due to the client's connection being disconnected during a large file/data being written to the response at the same time.

`business/web/v1/mid/errors.go` doesn't seem like a good place because of the following:
https://github.com/ardanlabs/service/blob/master/business/web/v1/mid/errors.go#L63-L66
will then call:
https://github.com/ardanlabs/service/blob/master/foundation/web/response.go#L39-L42
which ends up writing to the `http.ResponseWriter` anyways. Hence, when the call stack returns, the common denominator is to have `web.go` handle the error.

This will prevent what would currently cause a signaling of a "shutdown" and causing and outage, and instead, we now catch the error and stop the signaling of the shutdown. We should be "fine" with approach because if a client has terminated its connection, the backend should continue running.

### DIFF
--- a/foundation/web/web.go
+++ b/foundation/web/web.go
@@ -106,15 +106,21 @@ func (a *App) Handle(method string, group string, path string, handler Handler, 
 
 				// Usually, you get the broken pipe error when you write to the connection after the
 				// RST (TCP RST Flag) is sent.
-				// For example, the client's TCP connection is broken and the server writes to the
-				// connection.
+				// The broken pipe is a TCP/IP error occurring when you write to a stream where the
+				// other end (the peer) has closed the underlying connection. The first write to the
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
This is cool. I am going to encapsulate this a bit, but thanks!!!

ardan-bkennedy:
What is nice is I talk about needing to do this in the class. I like the idea of this code being in the foundation and not business.

jtrinh27:
Thanks. I’m glad it’s something useful. I can’t wait to see how you can improve on this. I’m still in the process of learning.

PR #229, opened 2022-07-12
Update dockerfile.metrics

jruariveiro:
Remove dup copy command


ardan-bkennedy:
This was found by a previous PR and fixed. Thanks for reporting it.

PR #237, opened 2022-09-03
fix missing field user_id

shalimski:
in seed script field user_id is empty on table sales


ardan-bkennedy:
Funny this was missed, but I guess we never implemented this API.  Thanks!

Issue #236, opened 2022-08-30
defer tx.Rollback


oussamm:
On database.go line 115
we create a boolean `mustRollback := true`

I am not sure if this boolean is really needed.
This [doc ](https://go.dev/doc/database/execute-transactions) says:
Defer the transaction’s rollback. If the transaction succeeds, it will be committed before the function exits, making the deferred rollback call a no-op. If the transaction fails it won’t be committed, meaning that the rollback will be called as the function exits.

Are we doing this because of the tracing?

oussamm:
Oh I get it, it's because that tx.Rollback, will return an `ErrTxClosed` if the transaction is made.

ardan-bkennedy:
I read the document and assuming sqlx is using the same semantics, you could be right.  That being said, I do not like these semantics because it breaks the golden rule, "don't make things easy to do, make things easy to understand".

It would be confusing to see a commit and rollback being executed. If I saw this in code, it would pause a code review. Now we need comments to explain why this is ok.

I think I rather have the logic be clear that if commit is executed, we don't call rollback.

ardan-bkennedy:
I have thought about this for several days and I changed my mind. I looked at the code sqlx was calling and it was from the stdlib. I studied the code and saw what you saw. So I changed the code to defer the rollback and removed the extra logic with the flag. Thanks for pointing this out!!

oussamm:
My pleasure Bill, one last thing, we could do this:
``` Go
if err := tx.Rollback(); err != nil && !errors.Is(err, sql.ErrTxDone) {
		log.Errorw("unable to rollback tran", "trace_id", traceID, "ERROR", err)
}
```

PR #240, opened 2022-10-31
small fix tx rollback

oussamm:
following this issue: https://github.com/ardanlabs/service/issues/236


ardan-bkennedy:
I implemented this change. Thanks for pointing it out.

PR #244, opened 2022-11-18
update sales binary name

belsakn:
Hello, 

I'm just following along the course you guys made. And the binary names don't match. So I thought I just make a PR to fix it.
BTW interesting stuff and nice work with the repo 👍


ardan-bkennedy:
This is cool. Thanks

PR #249, opened 2023-01-25
changing http status 500 to 502 as it is not application error itself.

halilylm:
As there is connection error to db, it is not application error, it is a connection error to another application. 502 would be better choice here.


nstogner:
Would a 502 be appropriate here since the service is not really acting as a proxy to the gateway? Seems like the failure of the db connection is more of an issue internal to the workings of the service?

PR #250, opened 2023-01-25
Fixing metric collector http client

halilylm:
DualStack is deprecated as Fast Fallback is enabled by default.


ardan-bkennedy:
Can you rollback the 502 change so I can merge this?

ardan-bkennedy:
I pushed the change manually. Thanks!

PR #251, opened 2023-01-25
Fixing productgrp handler filter method

halilylm:
ByQuantity is needed to be used instead of ByCost.


ardan-bkennedy:
Any chance these can be separate changes?

ardan-bkennedy:
I pushed the change manually. Thanks!

PR #253, opened 2023-02-03
avoid hanging on pinging

halilylm:
hello,
if there is no deadline in parent context, it will hang on pinging process, I will add extra check. If there is no context deadline, One second given by default.

### DIFF
--- a/business/sys/database/database.go
+++ b/business/sys/database/database.go
@@ -78,6 +78,15 @@ func Open(cfg Config) (*sqlx.DB, error) {
 // StatusCheck returns nil if it can successfully talk to the database. It
 // returns a non-nil error otherwise.
 func StatusCheck(ctx context.Context, db *sqlx.DB) error {
+	// If there is no timeout in parent context
+	// stop using hanging out on ping.
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(time.Second)
+	}
+	ctx, cancel := context.WithDeadline(ctx, deadline)
+	defer cancel()
+
 	var pingError error
 	for attempts := 1; ; attempts++ {
 		pingError = db.Ping()


ardan-bkennedy:
You are right and I am not sure I want to do this defensive coding right now. This kind of stuff adds 20%+ more code. It is the responsibility of the caller to pass a deadline.

ardan-bkennedy:
In the end, we provide a facility to allow this to timeout. If the caller doesn't want to provide that value, it's on them. It's been my experience this is the best practice with context.

halilylm:
Yes, I generally leave it to caller but when backing off to some external source, for loop hurting me, so I do it so, I thought it was appropriate to use it here. Thanks.

ardan-bkennedy:
I would ask why would a developer not provide a timeout? Why would we not catch this in a code review?

business/sys/database/database.go
```diff
@@ -78,6 +78,15 @@ func Open(cfg Config) (*sqlx.DB, error) {
 // StatusCheck returns nil if it can successfully talk to the database. It
 // returns a non-nil error otherwise.
 func StatusCheck(ctx context.Context, db *sqlx.DB) error {
+	// If there is no timeout in parent context
```
ardan-bkennedy, line 81: Line space above a comment and proper sentence structure. This comment could probably be all on one line.

business/sys/database/database.go
```diff
@@ -78,6 +78,15 @@ func Open(cfg Config) (*sqlx.DB, error) {
...
+	// stop using hanging out on ping.
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(time.Second)
+	}
+	ctx, cancel := context.WithDeadline(ctx, deadline)
```
ardan-bkennedy, line 87: Why not use content.WithTimeout and do everything in the if statement

```
if !ok {
   ctx, cancel = context.WithTimeout(ctx, time.Second)
   defer cancel()
}

**Comment by ardan-bkennedy** [comment]
It might be nice to keep for educational purpose. To explain the philosophy, but also show how you could dectect an empty deadlined context.

File: business/sys/database/database.go (hunk 2 above)
Comment (ardan-bkennedy) [reply business/sys/database/database.go:87]: This syntax won't work

Comment (ardan-bkennedy) [reply business/sys/database/database.go:87]:
```
if _, ok := ctx.Deadline(); !ok {
	var cancel context.CancelFunc
	ctx, cancel = context.WithTimeout(ctx, time.Second)
	defer cancel()
}
```

ardan-bkennedy: That reduced the logic down to 4 lines and only touched the context if there is no deadline. What do you think?

ardan-bkennedy:
I merged this change into master from my other branch!

business/sys/database/database.go
halilylm: It looks better and shorter than my implementation. Sorry being late to respond.

halilylm:
What about refactoring https://github.com/ardanlabs/service/blob/e383d2f70ead2c9eb1895b0ee5abe6f47a46ec91/foundation/worker/worker.go#L106 this part as well as we do here?

ardan-bkennedy:
Wow, I don't remember writing this, but how much cleaner is that code!

ardan-bkennedy:
In this case I needed a cancel function and I wanted to keep any user cancellation timeout.

ardan-bkennedy:
I think the logic we are using in StaticCheck is the right logic.

halilylm:
I went through all source code line by line :smile: I am thrilled by ultimate go training you give so I wanted to see how you decide things when writing web application.

ardan-bkennedy:
Thanks for the help and the extra eyes. Have you taken the live class?

I added better suport for Roles and cleaned some code this evening.

halilylm:
No I didn't know it. I am looking forward to watching Blockchain live classes on 13, 14 Feb.

ardan-bkennedy:
Send me an email.  bill@ardanlabs.com

halilylm:
I am sending. I also just sent an email for golang bridge for picking an program.  :smile:

PR #258, opened 2023-03-25
remove container leak

halilylm:
when starting a container, for example redis, if there is any typo when passing port argument, StartContainer will produce an error to caller however this image is running in the background we just couldn't extract ip, so as we created this container successfully, we should stop it when an error is happening after container started.


CONTRIBUTORS
```diff
@@ -61,3 +61,4 @@ Wietse Venema <wietsevenema@gmail.com>
 William Kennedy <bill@ardanlabs.com>
 Wyatt Johnson <wyattjoh@gmail.com>
 Zachary Johnson <zachjohnsondev@gmail.com>
+Halil İbrahim Yıldırım <ihalil95@gmail.com>
```
ardan-bkennedy, line 64: This needs to be in Alphabetical order please.

PR #270, opened 2023-05-14
Minimize or mitigate the chance of lock contention in worker.go

deckarep:
- Small tweak to use a `read lock` when checking if a work key exists in the `running` map.
- Added a `trackWork` helper method to minimize the locked lifetime which can possibly introduce contention when synchronizing a shutdown for a worker that takes awhile to finish.


ardan-bkennedy:
Much Better

PR #276, opened 2023-06-06
Add the log.Logger wrapper

dlsniper:
This adds back the error logging for the http.Server


ardan-bkennedy:
I'm going to accept this but make some changes. I missed that API.

PR #282, opened 2023-07-24
Fix foundation.app.handle otel span

levikobi:
Hi @ardan-bkennedy!

I noticed a small issue regarding the OpenTelemetry span management in the web framework:

Currently, the span is being created before there's a tracer in the context. As a result, the AddSpan only creates a new span, without actually starting it and attaching the received attributes. So, the solution is to construct the span and ctx before calling the AddSpan (similar to what happened before).

I also added an endpoint attribute, just thought it could be helpful.

### DIFF
--- a/foundation/web/web.go
+++ b/foundation/web/web.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/dimfeld/httptreemux/v5"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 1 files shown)


foundation/web/web.go
```diff
@@ -83,15 +84,18 @@ func (a *App) Handle(method string, group string, path string, handler Handler,
 // to the application server mux.
 func (a *App) handle(method string, group string, path string, handler Handler) {
 	h := func(w http.ResponseWriter, r *http.Request) {
-		ctx, span := AddSpan(r.Context(), "foundation.app.handle")
-		defer span.End()
+		ctx := r.Context()
+		span := trace.SpanFromContext(ctx)
 
-		ctx = SetValues(ctx, &Values{
+		ctx = SetValues(r.Context(), &Values{
```
ardan-bkennedy, line 90: This needs to be `SetValues(ctx`

ardan-bkennedy: I need @dlsniper to look at this. I think he identified that AddSpan would look at the context. I am not sure.

ardan-bkennedy:
I will accept this and then review and correct anything I think needs to.

ardan-bkennedy:
@levikobi this was a good change! Thanks!

levikobi:
> @levikobi this was a good change! Thanks!

I'm glad I could help :)

PR #283, opened 2023-07-27
Update makefile

clichi2002:
in case the helm does not finish in time which happened in my case


ardan-bkennedy:
Thanks! Can you add yourself to the CONTRIBUTORS file please.

PR #285, opened 2023-07-28
Update dev-database.yaml updating pg_hba.conf to allow connection to postgresql server

clichi2002:
because connection from pgcli was rejected from 
"clichi@ubuntu-22:~/GO_Course_Ardan_labs/service4.1-video$ pgcli postgresql://postgres:postgres@localhost connection failed: FATAL:  no pg_hba.conf entry for host "172.18.0.1", user "postgres", database "postgres", no encryption" there for pg_hba.conf needs to be modified to allow connection from 0.0.0.0/0 on our case being a dev local one way to fix that was to modify the location adding the args: ['-c', 'hba_file=/etc/pg_hba.conf'] to container start up arguments and create a config map with new file and mount that 
 volumes:
        - name: data
          persistentVolumeClaim:
            claimName: database-data
        - name: config
          configMap:
            name: pghbaconf
              #defaultMode: 0400
            items:
              - key: "pg_hba.conf" path: "pg_hba.conf" containers: - name: postgres image: postgres:15.3 volumeMounts: - name: data mountPath: /var/lib/postgresql/data - name: config readOnly: false mountPath: "/etc/pg_hba.conf" subPath: "pg_hba.conf"

Thanks

### DIFF
--- a/zarf/k8s/dev/database/dev-database.yaml
+++ b/zarf/k8s/dev/database/dev-database.yaml
@@ -37,17 +37,29 @@ spec:
         - name: data
           persistentVolumeClaim:
             claimName: database-data
+        - name: config
+          configMap:
+            name: pghbaconf
+              #defaultMode: 0400
+            items:
+              - key: "pg_hba.conf"
+                path: "pg_hba.conf"
       containers:
       - name: postgres
         image: postgres:15.3
         volumeMounts:
           - name: data
             mountPath: /var/lib/postgresql/data
+          - name: config
+            readOnly: false
+            mountPath: "/etc/pg_hba.conf"
+            subPath: "pg_hba.conf"
         resources:
           requests:
             cpu: "500m" # I need access to 1/2 core on the node.
           limits:
             cpu: "500m" # Execute instructions 50ms/100ms on my 1/2 core.
+        args: ['-c', 'hba_file=/etc/pg_hba.conf']
         env:
         - name: POSTGRES_PASSWORD
           value: postgres
@@ -88,3 +100,23 @@ spec:
   - name: postgres
     port: 5432
     targetPort: postgres
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pghbaconf
+  namespace: sales-system
+data:
+  # file-like keys
+  pg_hba.conf: |
+    local   all             all                                     trust
+    # IPv4 local connections:
+    host    all             all             0.0.0.0/0                trust
+    # IPv6 local connections:
+    host    all             all             ::1/128                 trust
+    # Allow replication connections from localhost, by a user with the
+    # replication privilege.
+    local   replication     all                                     trust
+    host    replication     all             0.0.0.0/0               trust
+    host    replication     all             ::1/128                 trust
+


ardan-bkennedy:
I have not needed this, nor students. So can you explain your situation here. What does this configuation file provide? I am guessing it exists in the container?

ardan-bkennedy:
I found this document.

https://www.postgresql.org/docs/current/auth-pg-hba-conf.html

ardan-bkennedy:
Hmmm. Do we know why you needed to set this and I don't nor anyone else I know?

dlsniper:
I wonder what's the environment difference.
For me it works out of the box, without the PR.

The changes between the shipped hba.conf and the proposed ones are small:

Screenshot_20230728_173542

ardan-bkennedy:
We need more information from @clichi2002 before we accept this.

clichi2002:
Hi

Background:

Ob 28 July 2023 I did run deploy the postgres on KIND and got the
connection refused error this is the image that have been downloaded
postgres                                   15.3      f0ff6ef79497   3 weeks
ago    412MB
If you are using the same image and you do not get the connection error
then I dont know.

In order to fix the connection error I did the modification of pg_hba.conf
that you see and pgcli worked.

Please try with the latest postgres image if you dont get pgcli connection
error I can not explain.

In case you have a similar connection issue now you have a fix.

PS Bill Thank you for the course. It is FANTASTIC!!!

clichi2002:
Hi

Furthermore
I have created today the KIND cluster from service directory on 2 different
laptops both running ubuntu one 5.19.0-50-generic and the second ubuntu-22
6.2.0-25-generic and the 5.19.0-50-generic postgresql accepted connection
without the change on the Linux ubuntu-22 6.2.0-25-generic I have to change
the zarf/k8s/dev/database/dev-database.yaml to get the connection with out
I got this error

***@***.***:~/GO_Course_Ardan_labs/service$ sudo make pgcli
pgcli
***@***.***
connection failed: FATAL:  no pg_hba.conf entry for host "10.244.0.5", user
"postgres", database "postgres", no encryption

I can not explain but this is the situation

Thanks

ardan-bkennedy:
Ok. I don't think it hurts to have this change if it's helping everyone. Let me test it and I'll merge. 

In the meantime, can you fix your other PR and put your name in alphabetical order please

ardan-bkennedy:
I added this manually to master.

clichi2002:
Hi

Last night I think I found the reason why this is happening.

In the dev-database.yaml you setup hostNetwork: true
 spec:
      dnsPolicy: ClusterFirstWithHostNet
      hostNetwork: true

Which means that the container running the postgresql  will use the network
stack of the machine rather than of the linux domain.

But we have a container within container as all the pods are running on one
docker container that is what KIND is.

So I believe in kernel 5 that would flow all the way to the
underlying network stack so the communication would happen over 127.0.0.1
in kernel 6 I think they fixed this security issue so container within
container can not escape

Using this unusual setting  hostNetwork: true I think this connection issue
has happened before that was a way to go around.

I dont have time to look at bug fixing in kernel 6 but I believe this is
one of those.

I hope that helps

Thanks

clichi2002:
Hi Bill

You have asked once why use namespace in kubernetes what is the advantages
One of them for example this config

Host         string
`conf:"default:database-service.sales-system.svc.cluster.local"`

If are in the same namespace for example sales-system-dev or
sales-system-prepod or sales-system-pod

It can be


Host         string `conf:"default:database-service"`


The way dns works inside kubernetes is the namespaces and the rest is
auto appended so you have the same configuration on all env just the
namespace differs

Another one is to cup the resources used by an namespace

Another one is network security if you implement network security with
network policies,

 It is very easy to restrict traffic between namespaces but by default
is allowed inside the same namespace.

Sorry it took so long but I am reviewing your code and it click that I
did not sent you that email

All the best

clichi2002:
Hi

Did it work?

ardan-bkennedy:
Yea, perfectly

clichi2002:
Great news

clichi2002:
Hi

BTW service6-video is a game changer fantastic work!
Thank you!

clichi2002:
Hi  Bill

One more thing about using namespace deleting the namespace will
delete all namespace objects.
You can do 'kubectl delete ns ns_name' and all the namespace object
will be deleted for example will delete pvc but not pv first one is
namespace based pv is not
This will work in immutable environment where you create a new
namespace on each git deploy and destroy the old one for example new
namespe_githash or namespace_ver
I am a big fun of immutable environments
Just a thought  not sure if this will help

Kind regards

PR #297, opened 2023-08-24
Upgrades CORS handling

unamashana:
* Return CORS headers on pre-flight requests (OPTIONS) for defined routes
  * Return 404 if an OPTIONS request comes in for an undefined ROUTE
* Return CORS headers on defined routes if any other HTTP method is sent (irrespective of the response status code)


ardan-bkennedy:
I accepted this code by entering it by hand!  This was awesome!

PR #300, opened 2023-08-29
Bruno/home

Bruno-10:


business/core/home/home.go
```diff
@@ -0,0 +1,203 @@
+package home
```
ardan-bkennedy, line 1: Missing package comment

business/core/home/home.go
```diff
@@ -0,0 +1,203 @@
...
+// retrieve data
+type Storer interface {
+	ExecuteUnderTransaction(tx transaction.Transaction) (Storer, error)
+	Create(ctx context.Context, nh Home) error
+	Update(ctx context.Context, hme Home) error
+	Delete(ctx context.Context, hme Home) error
+	Query(ctx context.Context, filter QueryFilter, orderBy order.By, pageNumber int, rowsPerPage int) ([]Home, error)
+	Count(ctx context.Context, filter QueryFilter) (int, error)
+	QueryByID(ctx context.Context, homeID uuid.UUID) (Home, error)
+	QueryByUserID(ctx context.Context, userID uuid.UUID) ([]Home, error)
```
ardan-bkennedy, line 33: This minics what we did in products so we are good. Just want to say this for the future. Anytime we bring back a collection we need to ask ourselves if we need paging. I don't expect a user to have thousands of homes, but if they did we could be in trouble.

business/core/home/home.go
```diff
@@ -0,0 +1,203 @@
...
+		evnCore: c.evnCore,
+		log:     c.log,
+	}
+
+	return c, nil
+}
+
+// Create adds a Home to the database. It returns the created Home with
+// fields like ID and DateCreated populated
+func (c *Core) Create(ctx context.Context, nh NewHome) (Home, error) {
```
ardan-bkennedy, line 89: I hate this comment and I know you copied it. Plus you are missing a closing period. 

We probably don't want to say database since the store abstraction might not be a database.

I think the second sentence is about us returning a complete Home with it's unique id and the date created.

We should fix this in all domains.

business/core/home/event.go
```diff
@@ -0,0 +1 @@
+package home
```
ardan-bkennedy, line 1: I am wondering if we should have this file if there are no events?

business/core/home/model.go
```diff
@@ -0,0 +1,55 @@
...
+	ID          uuid.UUID
+	Address     Address
+	Type        string
+	UserID      uuid.UUID
+	DateCreated time.Time
+	DateUpdated time.Time
+}
+
+// Address represents an individual address.
+type Address struct {
```
ardan-bkennedy, line 20: Put the Address type before Home since Home uses it. I like to think of the types as actors and when we can we should present them in the order of appearance.

business/core/home/stores/homedb/homedb.go
```diff
@@ -0,0 +1,216 @@
...
+
+// ExecuteUnderTransaction constructs a new Store value replacing the sqlx DB
+// value with a sqlx DB value that is currently inside a transaction.
+func (s *Store) ExecuteUnderTransaction(tx transaction.Transaction) (home.Storer, error) {
+	ec, err := db.GetExtContext(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	s = &Store{
```
ardan-bkennedy, line 40: Oh, this is bad and I know you copied it. We need to fix this. I want to construct with value semantics and use the & on the return. Let's fix that everywhere.

business/core/home/stores/homedb/homedb.go
```diff
@@ -0,0 +1,216 @@
...
+		db:  ec,
+	}
+
+	return s, nil
+}
+
+// Create adds a Home to the database. It returns an error if something went wrong
+func (s *Store) Create(ctx context.Context, hme home.Home) error {
+	const q = `
+      INSERT INTO homes
```
ardan-bkennedy, line 51: Check you tabbing in the editor.

business/core/home/stores/homedb/homedb.go
```diff
@@ -0,0 +1,216 @@
...
+// Query retrieves a list of existing homes from the database.
+func (s *Store) Query(ctx context.Context, filter home.QueryFilter, orderBy order.By, pageNumber int, rowsPerPage int) ([]home.Home, error) {
+	data := map[string]interface{}{
+		"offset":        (pageNumber - 1) * rowsPerPage,
+		"rows_per_page": rowsPerPage,
+	}
+
+	const q = `
+	SELECT
+		*
```
ardan-bkennedy, line 94: This * burned us a few times on TP. We should fix this everywhere and list out the fields.

business/core/home/stores/homedb/homedb.go
```diff
@@ -0,0 +1,216 @@
...
+
+	orderByClause, err := orderByClause(orderBy)
+	if err != nil {
+		return nil, err
+	}
+
+	buf.WriteString(orderByClause)
+	buf.WriteString(" OFFSET :offset ROWS FETCH NEXT :rows_per_page ROWS ONLY")
+
+	var dbHses []dbHome
```
ardan-bkennedy, line 109: I would argue this variable name is wrong.  dbHmes

business/core/home/stores/homedb/homedb.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Update modifies data about a Product. It will error if the specified ID is
+// invalid or does not reference an existing Product.
+func (s *Store) Update(ctx context.Context, hme home.Home) error {
+	const q = `
+	UPDATE
+		homes
+	SET
+    "address_1"     = :address_1,
```
ardan-bkennedy, line 126: Tabbing

business/core/home/stores/homedb/model.go
```diff
@@ -0,0 +1,28 @@
...
+
+import "github.com/ardanlabs/service/business/core/home"
+
+// dbHome represents an individual home.
+type dbHome struct{}
+
+// =============================================================================
+
+func toDBHome(hme home.Home) dbHome {
+	hmeDB := dbHome{}
```
ardan-bkennedy, line 11: You forgot to assign the fields.

business/core/home/stores/homedb/model.go
```diff
@@ -0,0 +1,28 @@
...
+
+	return hmeDB
+}
+
+func toCoreHome(dbHse dbHome) home.Home {
+	hme := home.Home{}
```
ardan-bkennedy, line 17: Forgot the assign the fields and check the variable name of the parameter.

business/core/home/event.go
Bruno-10: I created it and but I'm not sure how events work, so just left it empty, I'll delete it for the next PR

business/core/home/home.go
Bruno-10: I added paging in case we want to have an admin view of all homes. Let me know and I'll change it!

business/core/home/home.go
Bruno-10: Will do then. Comments seem weird all over the place, I'll fix whatever I found odd.

business/core/home/model.go
Bruno-10: I agree.

business/core/home/stores/homedb/homedb.go
Bruno-10: What kind of tabbing do you use? I have it on 2 spaces right now. Ooh the string might be not taken into account.

business/core/home/stores/homedb/homedb.go
Bruno-10: Will do that for all domains then.

business/core/home/stores/homedb/homedb.go
Bruno-10: Oh, I kept the old naming. Missed that completly sorry

business/core/home/stores/homedb/homedb.go
Bruno-10: Okay!

business/core/home/home.go
ardan-bkennedy: The Query call should have paging. I was talking about QueryByUserID since it also returns a collection. Don't change anything at this time.

business/core/home/event.go
ardan-bkennedy: Nah, leave the file for now. I left this comment to start a conversation. I think it's good to have it.

business/core/home/stores/homedb/homedb.go
ardan-bkennedy: Use 4 spaces per tab. That is the Go idiom.  That is not the problem here. I think you have some combination of tab and space with these queries.

PR #301, opened 2023-08-30
Bruno/home

Bruno-10:


business/core/home/home.go
```diff
@@ -0,0 +1,203 @@
...
+// Core manages the set of APIs for home api access.
+type Core struct {
+	log     *logger.Logger
+	usrCore *user.Core
+	evnCore *event.Core
+	storer  Storer
+}
+
+func NewCore(log *logger.Logger, evnCore *event.Core, usrCore *user.Core, storer Storer) *Core {
+	core := Core{
```
ardan-bkennedy, line 55: The variable name should be `c`

ardan-bkennedy: Let's fix the data semantics as well in product and user

PR #324, opened 2023-10-26
replaced dbsql/pq with dbsql/pgx

nhaancs:


ardan-bkennedy:
Wow. Can't believe I missed that

PR #325, opened 2023-10-27
HTTP client wrapper

nhaancs:
I added an HTTP client wrapper that supports proxy, tracing, and logging.


ardan-bkennedy:
Thanks for this but I'm not going to accept this. Vault is such a small part of this project and I don't want to bring in all this extra code.

nhaancs:
Thanks for your feedback. Just explain more about my case, my micro-service usually makes HTTP requests to other services, and the httpclient package helps me debug and monitor connections to those services easier. I added it to Vault just to demo how it can be used.

ardan-bkennedy:
I see. I think for your use case this is a good idea. This project needs to stay as MVP as possible. So tucking it inside vault just to have it doesn't work :(

THANKS!!

PR #332, opened 2023-11-27
Adding HTTP Shortcut Methods to Routing

josemukorivo:
This pull request adds convenient HTTP shortcut methods to the existing routing functionality.

Previously, routing was done using a longer syntax, such as:
`app.Handle(http.MethodGet, version, "/products", hdl.Query, authen, ruleAny)`
With this contribution, you can now use shorter and more expressive syntax like:
`app.Get(version, "/products", hdl.Query, authen, ruleAny)`

The motivation behind this change is to provide a more concise and expressive syntax for routing within the app. This not only enhances readability but also reduces boilerplate code when defining routes.


ardan-bkennedy:
Thank you for this but I can't accept this PR at this time. I do understand the change and think it's not a bad idea. For teaching purposes with the router I use, it would make things more difficult for me.

Plus, in go 1.22 I'm waiting to see what the new router looks like and will switch.

josemukorivo:
Thanks for taking the time to review the pull request. I completely understand your concerns about the impact on your teaching with the current router setup.

PR #353, opened 2024-02-12
Add auth header to make curl-create target

levikobi:
This request will not work without the authorization header

### DIFF
--- a/makefile
+++ b/makefile
@@ -461,7 +461,11 @@ live:
 	curl -il http://localhost:3000/v1/liveness
 
 curl-create:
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
If we are going to do this, are there other commands that can take advantage of this readability?

levikobi:
I guess `otel-test` could be reformated the same way. Let me know if you want me to switch it up too, I'll update the commit.

ardan-bkennedy:
Consistency is everything. So let's do this to anything that goes beyond 80 columns

ardan-bkennedy:
I will accept this and fix the other calls.

levikobi:
Sorry for the delay. I planned on pushing it later
If you want, I have pushed this for your reference https://github.com/levikobi/service/commit/0a0ed7bfbb23e4b05632083cdaaddb5c9b8ef943

ardan-bkennedy:
It's all good. I accepted the PR and cleaned up what I wanted. This was a nice change.

PR #358, opened 2024-03-06
feat: support podman

sudomateo:
This patch adds support for running service using Podman. There are a few notable changes that were required to enable Podman support, described below.

Podman keeps `HostIP` empty instead of using `0.0.0.0`. This behavior required an update to `extractIPPort`. More information in https://github.com/containers/podman/issues/17780.

Podman prefers to use fully-qualified container images (i.e., `example.com/foo/bar:latest`) but will perform [short-name aliasing](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md#short-name-aliasing) for unqualified container images. When an unqualified image such as `foo/bar:latest` is loaded into a kind cluster it will be stored as `localhost/foo/bar:latest`.

Using backticks for command substitution isn't POSIX compatible. Switched the backticks to `$()` instead to support alternative shells (e.g., `/bin/fish`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
           name: Install Go 1.22
           command: |
               sudo rm -rf /usr/local/go
-              wget -O go.tgz https://go.dev/dl/go1.22.0.linux-amd64.tar.gz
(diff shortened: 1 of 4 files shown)


ardan-bkennedy:
Thank you for this. I will review in the morning. If I have questions I'll reach out.

PR #366, opened 2024-03-24
remove seed database

himynamej:
do you think this change is good ,do we need seed database?
i want to do this in my project because its so complicated and must have different user base of every scenario


ardan-bkennedy:
The idea is to skip the seed. Not a bad idea. Do the test pass?

app/tooling/sales-admin/main.go
```diff
@@ -98,19 +98,6 @@ func processCommands(args conf.Args, log *logger.Logger, cfg config) error {
 			return fmt.Errorf("migrating database: %w", err)
 		}
 
-	case "seed":
```
ardan-bkennedy, line 101: You can't remove this from the admin. You still need the ability to do this for the project.

business/data/migrate/migrate.go
```diff
@@ -18,9 +16,6 @@ import (
 var (
 	//go:embed sql/migrate.sql
 	migrateDoc string
-
-	//go:embed sql/seed.sql
-	seedDoc string
```
ardan-bkennedy, line 23: Don't remove this. My project needs this for the production database and running locally. Just because you don't want it for testing doesn't make this not needed.

ardan-bkennedy:
I don't think you can remove the seeding because that adds an admin user to the database. This is needed to run tests. You don't know what order tests may run in. You might want to just run a home test. That is why the seeding adds two initial users to the system.

I am not sure you have tried running just a CreateProduct test, which won't work without users in the system.

ardan-bkennedy:
Granted, we don't need to query the seeded users in the app layer tests, but they are there and it can affect results. That is why I query them. You can create as many users as you want, and fix your tests to use that user.

ardan-bkennedy:
My last statement was wrong. You can create the user using the business layer API. The question is, should we use the seeded users since they exist, or always create a new user?

himynamej:
seed may cause confusion in writing tests

ardan-bkennedy:
I made changes to the repo based on your PR. I removed the seeding of the db during a test and removed those query calls. Let me know what you think.

himynamej:
its so good,thanks

ardan-bkennedy:
Thanks for the PR

PR #367, opened 2024-03-24
Remove TestGenerateSeed Funcs from business tests

Ali-Farhadnia:
I've made some changes to test functions in business layer for better readability. Here's why:

Consistency: I've used the same method for seeding data in all tests, which makes them easier to follow and maintain.

Clearer Tests: Putting the test data front  improves readability. It's easier to understand what each test is checking.

### DIFF
--- a/business/core/tests/home_test.go
+++ b/business/core/tests/home_test.go
@@ -22,6 +22,8 @@ func Test_Home(t *testing.T) {
 
 func homeCrud(t *testing.T) {
 	seed := func(ctx context.Context, userCore *user.Core, homeCore *home.Core) ([]home.Home, error) {
+		hmes := make([]home.Home, 1)
+
 		var filter user.QueryFilter
 		filter.WithName("Admin Gopher")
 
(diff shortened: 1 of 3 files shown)


ardan-bkennedy:
I should have used my test util functions for that. I see that now. I think it's better to be testing against random data.

Ali-Farhadnia:
Do you mean that util functions must  used for seeding the user too?

ardan-bkennedy:
I updated the code. All tests are using the core seed data functions. We don't want to hard code anything in the tests. If you have a special need for a DB record to look a certain way, you have the ability to do that. But I would use these Test functions as a first choice.

ardan-bkennedy:
MIssing something

Ali-Farhadnia:
thanks bill

ardan-bkennedy:
Look at it now. I think this is better. GREAT CATCH, I missed this :(

Ali-Farhadnia:
Thanks bill.Yes i think that is better now :)

PR #368, opened 2024-03-24
Fixing bug with sql

Ali-Farhadnia:
I found this bug when adding a test to check if the 'enabled' field is updated or not, but I cannot find a good way to test whether the user password and password confirmation fields have been updated.

### DIFF
--- a/business/core/tests/user_test.go
+++ b/business/core/tests/user_test.go
@@ -93,6 +93,8 @@ func userCrud(t *testing.T) {
 		Name:       dbtest.StringPointer("Jacob Walker"),
 		Email:      email,
 		Department: dbtest.StringPointer("development"),
+		Roles:      []user.Role{user.RoleUser},
+		Enabled:    dbtest.BoolPointer(false),
 	}
 
 	if _, err := api.Crud.User.Update(ctx, usrs[0], upd); err != nil {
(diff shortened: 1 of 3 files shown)


ardan-bkennedy:
I just started a road trip so I won't get to this for a day. You have two choices on the password. I think you get back the password hash, though I have to check. Assuming you do, you either check that it changed or you made a copy of the hash function, run the user:pass through it and check that it matches.

business/core/tests/user_test.go
```diff
@@ -127,6 +129,27 @@ func userCrud(t *testing.T) {
 		t.Errorf("Should be able to see updates to Department")
 	}
 
+	if len(saved.Roles) != len(upd.Roles) {
+		t.Logf("got: %v", saved.Roles)
+		t.Logf("exp: %v", upd.Roles)
+		t.Errorf("Should be able to see updates to Roles")
+	} else {
```
ardan-bkennedy, line 136: I HATE else clauses.  I am going to code this myself.  THIS WAS A GOOD FIND. Can't believe that has been missing.

ardan-bkennedy: Since we are only working with a single role in this slice, this will work fine.

```
if !saved.Roles[0].Equal(upd.Roles[0]) {
	t.Logf("got: %v", saved.Roles[0].Name())
	t.Logf("exp: %v", upd.Roles[0].Name())
	t.Errorf("Should be able to see updates to Role")
}
```

ardan-bkennedy:
I pushed a change for this, please check it out. If it's good please close the PR.  GREAT CATCH!

Ali-Farhadnia:
That's great.

PR #369, opened 2024-03-27
clean up model_test

himynamej:
hi maybe you forgot this.


ardan-bkennedy:
Nice catch. I'll fix that later today.

PR #371, opened 2024-03-28
remove Pass rom token

himynamej:
do you think we can remove pass from token?


ardan-bkennedy:
You are right, we don't need it. Nice!

PR #373, opened 2024-03-29
Question: Is it safe to use float for cost?

Ali-Farhadnia:
I have used decimal.Decimal ([github.com/shopspring/decimal](https://github.com/shopspring/decimal)) for cost calculations and it was ok for me.


ardan-bkennedy:
If float was not going to work, your next best solution is to use an int and keep everything in pennies. I don't that is important for this project.

PR #383, opened 2024-04-28
Fix make dev-gotooling

loadkpi:
Hello, thanks for this starter-kit!

It seems there are typos in `make dev-gotooling` command causing such error:

```
go: honnef.co/go/tools/programs/staticcheck@latest: module honnef.co/go/tools@latest found (v0.4.7), but does not contain package honnef.co/go/tools/programs/staticcheck
make: *** [makefile:141: dev-gotooling] Error 1
```


ardan-bkennedy:
Weird this has been wrong :(.  Thanks

PR #395, opened 2024-06-07
Correct makefile installation of watch

Shutt90:
Incorrect syntax in makefile for brew install watch - previously (brew instal watch)


Shutt90:
Closed as works with instal

PR #417, opened 2024-08-15
Do not return panic trace in the error message

pilowl:
When returning `errs.Internal` type of error, whole stack trace gets dumped in the error message. 
This does not seem to me like something intended, so I changed it to a `errs.InternalOnlyLog`, so that trace is only logged, and the api response message contains just `Internal Server Error` message in case if panic occurs.


ardan-bkennedy:
UGH. I forgot to do this when I added the support. I'm just going to make the change. THANK YOU

PR #425, opened 2024-09-13
fix Access-Control-Allow-Origin header

luxcgo:
CORS format only allow one Access-Control-Allow-Origin header

Reference
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin


ardan-bkennedy:
This was super interesting! Thanks for this fix.

PR #426, opened 2024-09-13
optimise cors response

luxcgo:


foundation/web/web.go
```diff
@@ -100,6 +100,7 @@ func (a *App) corsHandler(webHandler HandlerFunc) HandlerFunc {
 		for _, origin := range a.origins {
 			if origin == "*" || origin == reqOrigin {
 				w.Header().Set("Access-Control-Allow-Origin", origin)
+				break
```
ardan-bkennedy, line 103: I was looking at this more. I do except a collection of origins. I'm not really up to speed on CORS so I am wondering why I ever did that? Maybe a.origins should not be a slice?

foundation/web/web.go
```diff
@@ -76,7 +76,7 @@ func (a *App) EnableCORS(origins []string) {
 	a.origins = origins
 
 	handler := func(ctx context.Context, r *http.Request) Encoder {
-		return cors{Status: "OK"}
+		return nil
```
ardan-bkennedy, line 79: Do you think it's better to return a 204, which this will do?

foundation/web/web.go
ardan-bkennedy: Also, pull the latest code because I added some comments.

foundation/web/web.go
ardan-bkennedy: Yea, give the developer the option to have a list. OK

ardan-bkennedy: Ok cool!  Can you pull in the latest and then I will merge this. THANK YOU

ardan-bkennedy:
If you can think of a better way to provide this support within the scope of the web package I'm all ears. I appreciate the fixes.  When I switched to the new handler signature it added some wrenches in the mix. Everything is a tradeoff.

Heads up, for most of my projects I merge the API and APP layer together. The APP layer handlers use the web.HandlerFunc function signatures. I do plan to add GRPC to this project so people can see the benefit of separating the API and APP layers.

ardan-bkennedy:
I'm sorry I removed that, but is was very fragile. I didn't implement the tooling properly.

My plan is to try and use an LLM for both docs and generating new domains based on a data model. I'm just underwater and have not been able to get to that. Maybe by the end of the year.

ardan-bkennedy:
If you want access to the latest version of the Ultimate Software Design class let me know. It might help you understand the code and some of the decisions better.  It seems you've got a strong handle on things.

ardan-bkennedy:
Send me an email to bill@ardanlabs.com and I will get you that coupon code.

PR #427, opened 2024-09-14
fix(auth/main.go):change auth debug port to 6010

himynamej:


ardan-bkennedy:
Why?

himynamej:
auth debug port in k8s service is 6010

ardan-bkennedy:
Ok, good catch

PR #442, opened 2024-12-04
embed all static files

gradientsearch:
This change will embed static files starting with `.` and `_`. 

From the go [docs](https://pkg.go.dev/embed#hdr-Directives)

> If a pattern begins with the prefix ‘all:’, then the rule for walking directories is changed to include those files beginning with ‘.’ or ‘_’. For example, ‘all:image’ embeds both ‘image/.tempfile’ and ‘image/dir/.tempfile’.

### DIFF
--- a/api/services/sales/main.go
+++ b/api/services/sales/main.go
@@ -29,7 +29,7 @@ import (
 	Need to figure out timeouts for http service.
 */
 
-//go:embed static
+//go:embed all:static
 var static embed.FS
 
 var build = "develop"


ardan-bkennedy:
Why do we need this change?

angrieralien:
This would benefit users that use the `gonew` command to create their own version of the service and use a web framework that contains directories or files starting with `.` or `_` in the build output. I used svelte kit to build a simple static site and the build output contained a `_app` directory. The service returned 404s with the files existing in the `static` directory. This change would prevent that from happening.

ardan-bkennedy:
This code doesn't need this change. There may be a project that does in the future, but not this one. So making this change would technically be wrong. It doesn't change or affect the code in any way.

PR #448, opened 2024-12-23
Move queryparams

ameghdadian:
Moved the `queryparams` struct to be next to the `parseQueryParams` function to maintain consistency with other domains in the app layer.


mrbardia72:
i think , it seems good

PR #449, opened 2024-12-23
Error Handling in Validate Function

ameghdadian:
In the first commit, I used the newly added `Add` method of `FieldErrors` to append errors to the collection.

I'm not sure if I'm doing the right thing in the second commit. Since we're using the result of `Validate` function inside `web.Decode` function, and eventually this result is used at the app layer&mdash;like what comes next&mdash; I thought we could've wrapped any intermediate errors  and only then at the `app` layer, put them inside an `errs.Error`:

```go
       func (a *app) create(ctx context.Context, r *http.Request) web.Encoder {
	    ...
	    
	    if err := web.Decode(r, &app); err != nil {
	    	    return errs.New(errs.InvalidArgument, err)
	    }
	    
	    ...
        }
```

### DIFF
--- a/app/domain/homeapp/model.go
+++ b/app/domain/homeapp/model.go
@@ -91,7 +91,7 @@ func (app *NewHome) Decode(data []byte) error {
 // Validate checks if the data in the model is considered clean.
 func (app NewHome) Validate() error {
 	if err := errs.Check(app); err != nil {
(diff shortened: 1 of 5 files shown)


ardan-bkennedy:
This looks right. I will merge it and check out the tests

Issue #443, opened 2024-12-08
FieldErrors rendered as "Internal Server Error" 


ameghdadian:
`FieldErrors` results in 500 status code being rendered for two reasons:

1. `Errors` middleware explicitly looks for `errs.Error` type and if not found, returns 500 status code,
2.  `httpStatus` interface is not yet implemented for `FieldErrors`, therefore `case error:` switch case inside `web.Response` wins and results in `Internal Server Error`.

ardan-bkennedy:
You are right if a raw FieldErrors is returned, but that actually never happens.

```
func (a *app) create(ctx context.Context, r *http.Request) web.Encoder {
	var app NewHome
	if err := web.Decode(r, &app); err != nil {
		return errs.New(errs.InvalidArgument, err)
	}
```

We always convert errors into errs.Error in the app layer. So if this failed you would get the following under a 400.

```
validate: [{\"field\":\"type\",\"error\":\"type is a required field\"},{\"field\":\"address1\",\"error\":\"address1 is a required field\"},{\"field\":\"zipCode\",\"error\":\"zipCode is a required field\"},{\"field\":\"city\",\"error\":\"city is a required field\"},{\"field\":\"state\",\"error\":\"state is a required field\"},{\"field\":\"country\",\"error\":\"country is a required field\"}]
```

This is why you don't need the interfaces being implemented or any reference of this error type in the middleware.

ardan-bkennedy:
IMHO, you want to catch any error that is not an errs.Error flowing into the middleware and web package. You want to control everything running through the system. If you found something get through, you would want to fix it.

ameghdadian:
Thank you Bill.

So do we need to modify what's returned from `query` handlers in `service` project, as they're returning `FieldErrors` in some scenarios. Here's two instances where this happens: `[User App layer](https://github.com/ardanlabs/service/blob/cbf46d058a05e8bbb0892ba76f71de8fe0be6749/app/domain/userapp/userapp.go#L111) or [Product App layer](https://github.com/ardanlabs/service/blob/cbf46d058a05e8bbb0892ba76f71de8fe0be6749/app/domain/productapp/productapp.go#L83).
```go
func (a *app) query(ctx context.Context, r *http.Request) web.Encoder {
	qp, err := parseQueryParams(r)
	if err != nil {
		return errs.New(errs.InvalidArgument, err)
	}

	page, err := page.Parse(qp.Page, qp.Rows)
	if err != nil {
[...]

PR #457, opened 2025-01-19
Increase Test Coverage

ameghdadian:
Added missing tests to cover 400 scenarios when given malformed query params.
This is originally mentioned at [#443](https://github.com/ardanlabs/service/issues/443#issuecomment-2541594859):

> We need some tests to cover query filters. I though I had them so when the tests passed I thought I was good. What a classic mistake :(


ardan-bkennedy:
This is awesome!!!

PR #458, opened 2025-01-22
fix(queryParams): change "row" to "rows" in query params

abolfazlamini:
There was an issue with `make load` command which rows=2 not worked properly and returned 10 rows default.

### DIFF
--- a/app/domain/homeapp/filter.go
+++ b/app/domain/homeapp/filter.go
@@ -26,7 +26,7 @@ func parseQueryParams(r *http.Request) queryParams {
 
 	filter := queryParams{
 		Page:             values.Get("page"),
-		Rows:             values.Get("row"),
+		Rows:             values.Get("rows"),
 		OrderBy:          values.Get("orderBy"),
 		ID:               values.Get("home_id"),
 		UserID:           values.Get("user_id"),
--- a/app/domain/productapp/filter.go
+++ b/app/domain/productapp/filter.go
@@ -25,7 +25,7 @@ func parseQueryParams(r *http.Request) queryParams {
 
 	filter := queryParams{
 		Page:     values.Get("page"),
-		Rows:     values.Get("row"),
+		Rows:     values.Get("rows"),
 		OrderBy:  values.Get("orderBy"),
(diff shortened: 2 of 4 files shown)


ameghdadian:
> Why not fix the makefile? I'm shocked the tests didn't fail. I need to see what I was doing there first.

It's mainly because we use the same `RowsPerPage` value(i.e. 10) in tests as we're setting a default for `rows` value: https://github.com/ardanlabs/service/blob/fa83c1028a2e4b9653ad27099731b97ea069ca71/business/sdk/page/page.go#L26

Therefore inside this function:
```go
func parseQueryParams(r *http.Request) (queryParams, error) {
	values := r.URL.Query()

	filter := queryParams{
		Page:             values.Get("page"),
		Rows:             values.Get("row"),
		OrderBy:          values.Get("orderBy"),
		ID:               values.Get("user_id"),
		Name:             values.Get("name"),
		Email:            values.Get("email"),
		StartCreatedDate: values.Get("start_created_date"),
		EndCreatedDate:   values.Get("end_created_date"),
	}

	return filter, nil
}
```
`row` query param is not found and an empty string is passed to `page.Parse` function, leading to always using the default(i.e. 10) value.

ardan-bkennedy:
This was a great catch. My guess is I went with the singular for both for consistency. So I would rather fix the tests and makefile to use `row` instead of `rows`.

ardan-bkennedy:
I pushed a change to keep `row` as the query param and to fix the tests and makefile.

abolfazlamini:
> Why not fix the makefile? I'm shocked the tests didn't fail. I need to see what I was doing there first.

Cause the name of the field is Rows. ` Rows:             values.Get("..."),` 
and it is meaningful for pagination(number of rows). `row` is ambiguous which pretending to selecting an specific row!

ardan-bkennedy:
SHIT, you are right. Let's change that back to rows.

ardan-bkennedy:
Ok, I made the change. Sorry, you were right.

abolfazlamini:
> Ok, I made the change. Sorry, you were right.

Great, Thank you for your consideration.

PR #462, opened 2025-02-24
Select priority issue and missing dot.

yasinErdemli:
In the comment of the code it says shutdown is a priority, but in select statement the order is random. There is also a missing dot.


ardan-bkennedy:
I would not change the select the way you did. The G needs to handle all three at the same time.

Thanks for finding the period

PR #465, opened 2025-03-02
small fix to errors.New

ousloob:


ardan-bkennedy:
Not sure these are important changes. Tell me why you want to make them.

api/services/auth/main.go
```diff
@@ -167,7 +167,7 @@ func run(ctx context.Context, log *logger.Logger) error {
 	}
 
 	if n1+n2 == 0 {
```
ousloob, line 169: If there is nothing loaded, the return will wrap the error of `ks.LoadByFileSystem(os.DirFS(cfg.Auth.KeysFolder))`
which doesn't make sense since this error will be nil

ardan-bkennedy: That logic is there because there are 2 options for the keys. When I run the project locally we use the LoadByFile since we have a private key file. When we are in production we don't use that. We add the key to an Env var as a secret. What that logic is for it to determine if we had a key at all. Since the API returns the number of keys it finds, 0 would mean no keys loaded.

ousloob: For other cases it's just matter of best practice, we use `errors.New` for static errors and `fmt.Errorf` in case of messages formatting wrapping or dynamic content.

ardan-bkennedy (approved):

PR #522, opened 2025-06-12
Refactor: Logger Discard

ameghdadian:


ardan-bkennedy:
You were right on that. I totally spaced, thank you!

PR #524, opened 2025-07-07
Make frontend great again!

megaes:
Frontend is quite buggy, but it doesn't work out of the box at the moment. 
This bunch of short commits make it alive.


ardan-bkennedy:
Thanks for this. I am not a FE and was close to removing this from the project.

PR #527, opened 2025-07-12
Refactor sampler to use combined `url.path` and `url.query` for endpoint matching.

megaes:
HTTP semantic conventions changed.
http.target attribute was split into two attributes, url.path and url.query.

https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-spans.md#http-client-server-example

### DIFF
--- a/foundation/otel/sampler.go
+++ b/foundation/otel/sampler.go
@@ -1,6 +1,8 @@
 package otel
 
 import (
+	"errors"
+
 	"go.opentelemetry.io/otel/sdk/trace"
 )
 
@@ -16,7 +18,7 @@ func newEndpointExcluder(endpoints map[string]struct{}, probability float64) end
 	}
 }
 
-func endpoint(parameters trace.SamplingParameters) string {
+func endpoint(parameters trace.SamplingParameters) (string, error) {
 	var path, query string
 
 	for _, attr := range parameters.Attributes {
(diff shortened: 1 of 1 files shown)


foundation/otel/sampler.go
```diff
@@ -16,14 +16,31 @@ func newEndpointExcluder(endpoints map[string]struct{}, probability float64) end
...
+		switch attr.Key {
+		case "url.path":
+			path = attr.Value.AsString()
+		case "url.query":
+			query = attr.Value.AsString()
+		}
+	}
+
+	if path != "" && query != "" {
+		path += "?" + query
```
ardan-bkennedy, line 32: We should return the value here and then return "" at the end. When you write code like this where the return is not explict, it requires a lot of mental effort to understand what is going on.

I don't remember any of this code :(

ardan-bkennedy: I looked more and I see we are using it to remove liveness and readiness from the traces. So you are getting the route and then checking it. I can test this, but lets clean up the code first.

foundation/otel/sampler.go
```diff
@@ -16,14 +16,31 @@ func newEndpointExcluder(endpoints map[string]struct{}, probability float64) end
...
+		case "url.query":
+			query = attr.Value.AsString()
+		}
+	}
+
+	if query != "" && path != "" {
+		return path + "?" + query
+	}
+
+	return path
```
ardan-bkennedy, line 35: You want to return the literal value of "" here. When you use the path variable we have to read the code to know what is being returned

megaes: Mr. Kennedy, I am not a native English speaker—sorry in advance for any possible misunderstandings.

PR #529, opened 2025-07-12
fix: handle specific application OK error case in isError function

mannguyen0107:
This is a tiny improvement to the application errs. There is an error enum called `OK` and it just get log as error which is not right. More importantly if you have a transaction middleware and you return this error the transaction will get rollback which is not correct as well.

I could be using this wrong but this is the way I'm returning 200 OK without making a struct conforming to the `web.Encoder` interface.

### DIFF
--- a/app/sdk/mid/mid.go
+++ b/app/sdk/mid/mid.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 
 	"github.com/ardanlabs/service/app/sdk/auth"
+	"github.com/ardanlabs/service/app/sdk/errs"
 	"github.com/ardanlabs/service/business/domain/homebus"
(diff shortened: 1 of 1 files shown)


ardan-bkennedy:
I don't like how you are using an error to return a 200. That to me is a flaw in the system. I think I'm going to remove the errs.OK from the platform

ardan-bkennedy:
I am changing errs.OK to errs.None.  Thank you for the change to isError.

> I could be using this wrong but this is the way I'm returning 200 OK without making a struct conforming to the web.Encoder interface.

This means you are creating an errs.Error with errs.OK (now None) and returning an error document with status 200. That is super interesting and I didn't realize that could be done.

I've added your change manually.

PR #530, opened 2025-07-22
fix: eliminate nil receiver warnings in filter parseFilter functions Initialize FieldErrors as empty slice instead of nil across all domain filters

D-sense:
This PR fixes filters: initializes FieldErrors as empty slice to prevent nil receiver warnings

- Replace `var fieldErrors errs.FieldErrors` with `make(errs.FieldErrors, 0)`
- Update nil checks to use `len(fieldErrors) > 0` instead of `!= nil`
- Applied across auditapp, productapp, userapp, homeapp, and vproductapp filters
- Eliminates linter warnings about calling methods on potentially nil receivers

### DIFF
--- a/app/domain/auditapp/filter.go
+++ b/app/domain/auditapp/filter.go
@@ -44,7 +44,7 @@ func parseQueryParams(r *http.Request) (queryParams, error) {
 }
 
 func parseFilter(qp queryParams) (auditbus.QueryFilter, error) {
-	var fieldErrors errs.FieldErrors
+	var fieldErrors = make(errs.FieldErrors, 0)
 	var filter auditbus.QueryFilter
 
 	if qp.ObjID != "" {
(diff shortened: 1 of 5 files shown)


ardan-bkennedy:
Which linter are you using? I use staticcheck

ardan-bkennedy:
I would argue that unless you need an empty collection keep it nil. The logic we have would never attempt to access the nil slice at an index level. Plus an empty slice where one tries to access index 0 would also fail.  So using an empty slice here doesn't solve your argument. 

I never like changing code because a linter has defined some rule and can't do proper code analysis.

D-sense:
> I would argue that unless you need an empty collection keep it nil. The logic we have would never attempt to access the nil slice at an index level. Plus an empty slice where one tries to access index 0 would also fail. So using an empty slice here doesn't solve your argument.
> 
> I never like changing code because a linter has defined some rule and can't do proper code analysis.

Thanks for explaining that. I will close the pr.

PR #533, opened 2025-08-19
fix: Return NotFound instead of Unauthenticated for missing resources

Jkarage:
I was misunderstanding the error response I was getting from my cloned service. I just found this in the auth middleware—it seems intentional though. Is it?

### DIFF
--- a/app/sdk/mid/authorize.go
+++ b/app/sdk/mid/authorize.go
@@ -72,7 +72,7 @@ func AuthorizeUser(client *authclient.Client, userBus userbus.ExtBusiness, rule 
 				if err != nil {
 					switch {
 					case errors.Is(err, userbus.ErrNotFound):
-						return errs.New(errs.Unauthenticated, err)
+						return errs.New(errs.NotFound, err)
 					default:
 						return errs.Newf(errs.Unauthenticated, "querybyid: userID[%s]: %s", userID, err)
 					}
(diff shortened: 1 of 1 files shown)


Jkarage:
@ardan-bkennedy thanks,  My worry is we're failing with this, before we have checked if the current user has access or not. So the error is not totally authorization issue.

Also it might be confusing to get Unauthorized, and the error is product_id not found. 

What about checking the permissions before checking data in db.

ardan-bkennedy:
If the user doesn't exist, then you are not authorized. We are checking the user related to the request. The fact that the user doesn't exist is related to auth, not whether it exists or not. 

It would be a major security risk to tell the caller the user doesn't exist. That gives the caller information they shouldn't have.

Jkarage:
In the AuthorizeProduct middleware (just an example) even if I am Authorized to access this resource. If the particular `Product_id` am requesting/updating is not available in the database. I would still get 401.

ardan-bkennedy:
That to me is different. Auth isn't on a product, it's on a user. At that point you've passed security. Now the product you want doesn't exist.

ardan-bkennedy:
Hmmm we don't do a user check it seems on that. 

I need to look closer at that middleware. Technically I would think we need to verify the user as part of the Authorization.

give me a day or two to think about this. In my quick review I must be missing something.

maybe you found something here!

ardan-bkennedy:
I'm going to stick to my original comment. Any authorization middleware should fail with an Unauthenticated code. You really dont want to reveal anything else. I could aruge that since I am returning the error, that could be a security flaw.

PR #534, opened 2025-08-19
fix(gentoken): keystore load fallback

abolfazlamini:
**Changes:** Updates `api/tooling/admin/commands/gentoken.go`

**Summary:**
Fixes a bug where the `gentoken` command would crash immediately if the first keystore load method failed.

**What was fixed:**
- Added `os.Getenv("SALES_PEM")`
- Changed `SALAES_PEM` to `SALES_PEM` typo.

**Result:** The command is now more robust and will try all available methods to load the keystore before failing.

### DIFF
--- a/api/tooling/admin/commands/gentoken.go
+++ b/api/tooling/admin/commands/gentoken.go
@@ -43,7 +43,7 @@ func GenToken(log *logger.Logger, dbConfig sqldb.Config, keyPath string, userID 
 
 	ks := keystore.New()
 
-	n1, err := ks.LoadByJSON("SALAES_PEM")
+	n1, err := ks.LoadByJSON(os.Getenv("SALES_PEM"))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


abolfazlamini:
> This is a bad change. You want to retrieve config values in main and pass these into APIs. This hides and scatters config throughout the app.

No I do not want the whole point is the error on loading keys for auth. First time I encountered this bug in gentoken in admin tooling. Anyway the gentoken not perform at all, always return error cause of `"SALAES_PEM"`.

ardan-bkennedy:
I'm look at the code but pulling a value from the env is not the solution

ardan-bkennedy:
Ok. I finally got to look at the code. I completely understand the confusion and I am sorry. That config variable should have been `KeysJSON` where it contains the key information in a JSON format. Not some env var.

THANK YOU for seeing that. I have fixed the code.

PR #579, opened 2025-09-04
Delegate.Call error Propagation

ameghdadian:
I think it's a good idea to propagate callback function errors in `Delegate.Call` method upwards, so that we can rollback the transaction.

I'm also wondering, when inside one of these callback functions, is it a good practice to call into the same Business layer functionalities, or we should make use of functionalities provided in the Storage layer? I can see that `userBus.Authenticate` uses `userBus.QueryByEmail`, but I'm asking this to be sure.

Thanks

### DIFF
--- a/business/sdk/delegate/delegate.go
+++ b/business/sdk/delegate/delegate.go
@@ -4,6 +4,7 @@ package delegate
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/ardanlabs/service/foundation/logger"
 )
@@ -56,6 +57,7 @@ func (d *Delegate) Call(ctx context.Context, data Data) error {
 
 				if err := fn(ctx, data); err != nil {
 					d.log.Error(ctx, "delegate call", "err", err)
+					return fmt.Errorf("delegate: %w", err)
 				}
 			}
 		}


ameghdadian:
> I'm also wondering, when inside one of these callback functions, is it a good practice to call into the same Business layer functionalities, or we should make use of functionalities provided in the Storage layer? I can see that `userBus.Authenticate` uses `userBus.QueryByEmail`, but I'm asking this to be sure.
> 

Appreciate your comments when you get a chance!

ameghdadian:
I'd like to raise an issue. When using Query functions inside the callback functions (i.e. those invoked in `Delegate.Call`), we are constrained by the page size limit (100) defined in the `Parse` function of the `page` package.

Depending on the dataset size, we could add helper functions to either fetch data in chunks or retrieve the full dataset at once.

Would you like me to open a PR for this?

ardan-bkennedy:
I would question why you need to bring back so much data for some internal processing. That sounds like a smell to me.

In this case you probably have a specific thing you need to accomplish and it deserves it's own API.

ameghdadian:
> I would question why you need to bring back so much data for some internal processing. That sounds like a smell to me.
> 
> In this case you probably have a specific thing you need to accomplish and it deserves it's own API.

Oh, I see! Now I understand what I'm missing. 
Thank you!

> 3. Call the same API multiple times for the amount of data you need.
> 

Also, I prepared a function which helps to get the next page, whenever you need to repeatedly call the same `Query` API:

```go
package page

// ...

// Next calculates and returns the next page.
func Next(p Page, total int) iter.Seq2[Page, error] {
	return func(yield func(Page, error) bool) {
		maxPage := (total + p.rows - 1) / p.rows
		for i := p.number; i <= maxPage; i++ {
			rows := p.rows
			if i == maxPage && total%p.rows != 0 {
				rows = total % p.rows
			}

			next, err := Parse(fmt.Sprint(i), fmt.Sprint(rows))
			if !yield(next, err) {
				return
			}
		}
	}
}
```

I’m wondering if this is still handy for this project even though there's not a use case for it right now?

ameghdadian:
Sure!

It makes use of [iterators](https://go.dev/blog/range-functions#standard-push-iterators).

I tried to show how to use the function inside this [sandbox](https://go.dev/play/p/UaKpKhRry6I).

PR #619, opened 2025-12-01
Initial introduction of gRPC into the service repo

dlsniper:
This PR aims to introduce support for gRPC in the repository.
It's used under the authentication/authorization service as an alternative to the HTTP client.


zarf/proto/auth.proto
```diff
@@ -0,0 +1,67 @@
+syntax = "proto3";
```
tebeka: You might consider using 2023 edition and the [opaque API](https://go.dev/blog/protobuf-opaque)

tebeka: You'll need to update the code as well, `protoc` generates different code :)

PR #663, opened 2026-06-10
fixed audit default order by to use auditbus

Ali-Farhadnia:
# auditapp: use auditbus.DefaultOrderBy for the default sort

The audit query handler was falling back to `userbus.DefaultOrderBy`. Switched
it to `auditbus.DefaultOrderBy` and dropped the now-unused `userbus` import.

Only worked before because both order constants are `"a"`.


ardan-bkennedy:
WOW

PR #671, opened 2026-08-20
chore: modernize codebase for Go 1.26 stdlib features

hypergadam:
- errors.As with a declared target -> errors.AsType[T]
- proto.String/Int64 and dbtest pointer helpers -> new(v)
- reverse index loops -> slices.Backward
- counted loops -> for range N
- reflect.TypeOf((*T)(nil)).Elem() -> reflect.TypeFor[T]
- reflect.Ptr -> reflect.Pointer
- wg.Add/go/defer wg.Done -> wg.Go


ardan-bkennedy:
Nice and thanks

PR #672, opened 2026-08-29
chore: upgrade golang-jwt to v5

hypergadam:
- Bump github.com/golang-jwt/jwt to v5 in go.mod
- Update all import paths to use v5
- Update expected test error strings to match the new v5 parsing error format
- Run go mod vendor to sync dependencies


ardan-bkennedy:
This looks great. Thanks for the PR. Doing some final checks and we should be able to merge this.

